PR #12, opened 2021-07-20

VladyslavLekhter:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
+        for(User u : users){
+            if(u.getEmail().equals(email)){
+                return u;
```
Iryna-U, line 20:
```suggestion
        for(User user : users){
            if(u.getEmail().equals(email)){
                return user;
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if(email == null || password == null) return false;
+        if(UserService.findByEmail(email) == null) return false;
+        if(password == UserService.findByEmail(email).getPassword()){
+            return true;
+        }
         return false;
```
Iryna-U, line 19:
```suggestion
        User user = new UserService.findByEmail(email);
        if (email != null && password != null && user != null) {
        return user.getPassword().equals(password);
        }
        return false;
```

Iryna-U (changes requested):

PR #13, opened 2021-07-20

andrii-demchenko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService var = new UserService();
+        if (var.findByEmail(email) == null) {
+            return false;
+        } else {
+            return var.findByEmail(email).getPassword().equals(password);
+        }
```
Resci, line 19:
```suggestion
        UserService userService = new UserService();
        User user = userService.findByEmail(email);
        if (user == null) {
            return false;
        } else {
            return user.getPassword().equals(password);
        }
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (email.equals(users[i].getEmail())) {
+                return users[i];
+            }
+        }
```
Resci, line 22:
```suggestion
        for (int i = 0; i < users.length; i++) {
            if (users[i].getEmail().equals(email)) {
                return users[i];
            }
        }
```

Resci (changes requested):

Iryna-U (approved):

PR #21, opened 2021-08-02
implemented login and findByEmail methods, made findByEmail static

AlexRozv:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null) {
+            return userService.findByEmail(email).getPassword().equals(password);
```
Resci, line 16: Should we search the user inside the database twice?

Resci (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
AlexRozv: fixed

PR #22, opened 2021-08-02

Black3942:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = x.findByEmail(email);
+        if (y == null) {
+            return false;
+        } else {
+            return y.getPassword().equals(password);
```
Iryna-U, line 21:
```suggestion
        User user = userService.findByEmail(email);
        if (user == null) {
            return false;
        } else {
            return user.getPassword().equals(password);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user : users) {
+            if (email.equals(u.getEmail())) {
+                return u;
```
Iryna-U, line 20:
```suggestion
            if (user.getEmail().equals(email)) {
                return user;
```

Iryna-U (changes requested):

Iryna-U (changes requested):
Fix checkstyle

Iryna-U (approved):

PR #23, opened 2021-08-03
Add Login and FindByMail services

osa240:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@
...
      */
     public boolean login(String email, String password) {
-        return false;
+        boolean out = false;
+        UserService tempUserService = new UserService();
+        User tempUser = tempUserService.findByEmail(email);
+        if (tempUser != null) {
+            out = tempUser.getPassword().equals(password);
+        }
+        return out;
```
Iryna-U, line 22:
```suggestion
        UserService userService = new UserService();
        User user = userService.findByEmail(email);
        if (user != null) {
            return user.getPassword().equals(password);
        }
        return false;
```

Iryna-U (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
osa240: I fix it.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService tempUserService = new UserService();
```
Iryna-U, line 16:
```suggestion
        UserService userService = new UserService();
```

Iryna-U (approved):

PR #26, opened 2021-08-08
Complete the task HW-00

VladGural:
implemented a `login` method in `AuthenticationService` and `findByEmail` method in `UserService`

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,6 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = new UserService().findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        if (user.getPassword().equals(password)) {
+            return true;
+        }
```
Iryna-U, line 22:
```suggestion
        if (user != null) {
            return user.getPassword().equals(password);
        }
```

Iryna-U (changes requested):

Iryna-U (approved):

PR #27, opened 2021-08-09

dogovor:


Resci (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -2,15 +2,19 @@
...
+       * @param email    - user's email
+       * @param password - user's password
+       * @return true if user by email exists and passed password is equal to user's password.
+       * Return false in any other cases.
+       */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        return userService.findByEmail(email) != null
+                    && userService.findByEmail(email).getPassword().equals(password);
```
Iryna-U, line 17: Better create new variable - user
```suggestion
        User user = userService.findByEmail(email);
        return user != null && user.getPassword().equals(password);
```

Iryna-U (changes requested):

PR #29, opened 2021-08-10
 created login and findEmail method. All tests are passed 

netiazhevskyi:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
 package mate.academy.service;
 
+//import mate.academy.model.User;
```
Resci, line 3: Let's remove all comments

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null) {
+            return password.equals(userService.findByEmail(email).getPassword());
```
Resci, line 19: Should we search the user inside the database twice?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
```
Resci, line 6:
```suggestion
    private static final User[] users = new User[] {
```

Resci (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
netiazhevskyi: done, removed all comments

PR #38, opened 2021-08-11
implementation of user verification

jmenserg:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService check = new UserService();
+        if (check.findByEmail(email) != null
+                && check.findByEmail(email).getPassword().equals(password)) {
```

Iryna-U (changes requested):

PR #39, opened 2021-08-11
Adding methods in AuthenticationService and UserService

denysSmeiNewBom:


denyssmei:
GoGoGO

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
...
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (!(user == null) && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
Resci, line 22:
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
```
Resci, line 6:
```suggestion
    private static final User[] users = new User[] {
```

Resci (changes requested):

denyssmei:
Adding some changes in login method and User class

Iryna-U (approved):

PR #48, opened 2021-08-12
Hw-jv-oop

RostyslavF5R:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService user = new UserService();
+        if (user.findByEmail(email) != null
+                && user.findByEmail(email).getPassword().equals(password)) {
```

Iryna-U (approved):

PR #51, opened 2021-08-12
added solution

taras-oleksiuk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null) {
+            return password.equals(userService.findByEmail(email).getPassword());
```

Iryna-U (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
Iryna-U, line 16:
```suggestion
        User user = userService.findByEmail(email);
        if (user != null) {
            return user.getPassword().equals(password);
```

Iryna-U (changes requested):

PR #55, opened 2021-08-12
implemented task

Radikhovskiy:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User validUser = userService.findByEmail(email);
+        return password != null && validUser != null
+                && password.equals(validUser.getPassword());
```
Iryna-U, line 19: What happens if the password equals null?

Radikhovskiy: > What happens if the password equals null?
Method will return false
Should I rewrite it? Because there is no need to write <if - else> construction

Iryna-U (approved):

PR #57, opened 2021-08-12
added solution

MAFIOZNIK1999:
added solution


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,17 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
```
Iryna-U, line 6:
```suggestion
    private static final User[] users = new User[] {
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,17 @@
...
 
-    /**
-     * Find user by email. All users are stored in <code>private static final User[] users</code>
-     * @param email - the input parameter
-     * @return - user if his email is equal to passed email.
-     * Return <code>null</code> if there is no suitable user
-     */
     public User findByEmail(String email) {
+        for (User i : users) {
+            if (i.getEmail().equals(email)) {
```
Iryna-U, line 13:
```suggestion
        for (User user : users) {
            if (user.getEmail().equals(email)) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,12 @@
...
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
+
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null) {
+            return userService.findByEmail(email).getPassword().equals(password);
+        }
         return false;
```
Iryna-U, line 10: Should we search the user inside the database twice?
```suggestion
       User user = userService.findByEmail(email);
        if (user != null) {
            return user.getPassword().equals(password);
        } else {
            return false;
```

Iryna-U (changes requested):

PR #60, opened 2021-08-12
implemented a login  method in AuthenticationService and findByEmail …

ditroRV:
…method in UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = new UserService().findByEmail(email);
+        if ((user != null) && (user.getPassword() != null) && (user.getPassword().equals(password))) {
+            return true;
+        } else {
+            return false;
+        }
```
Iryna-U, line 21: this check will suffice
```suggestion
        if (user != null) {
            return user.getPassword().equals(password);
        }
        return false;
```

Iryna-U (changes requested):

Iryna-U (changes requested):
Fix checkstyle

Iryna-U (approved):

PR #61, opened 2021-08-12
login method in  AuthenticationService; findByEmail in UserService

SolnalarmJulia:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,30 @@
...
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (email == null || password == null) {
+            return false;
+        }
+        if (email.isEmpty() || password.isEmpty()) {
+            return false;
```
Iryna-U, line 21:
```suggestion
        if (email == null || password == null) {
            return false;
        }
``
this check will suffice

File: src/main/java/mate/academy/service/UserService.java [hunk 2]
```diff
@@ -3,18 +3,24 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    public static final User[] users = new User[]{
```
Iryna-U, line 6: Please, don't change the method signature
```suggestion
    private static final User[] users = new User[] {
```

Iryna-U (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,30 @@
...
+        }
+        UserService userService = new UserService();
+        User isUser = userService.findByEmail(email);
+        if (isUser == null) {
+            return false;
+        }
+        if (isUser.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
Iryna-U, line 28:
```suggestion
        User user = userService.findByEmail(email);
        if (user == null) {
            return false;
        }
        return user.getPassword().equals(password);
```

Iryna-U (changes requested):

PR #62, opened 2021-08-12
implementation of validation

jmenserg:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService check = new UserService();
+        User user = check.findByEmail(email);
```
Iryna-U, line 17:
```suggestion
        UserService userService = new UserService();
        User user = userService.findByEmail(email);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@
...
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
Iryna-U, line 21: Looks better
```suggestion
       
        return user != null && user.getPassword().equals(password);
```

Iryna-U (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
jmenserg: I understand. How can I change opened pull request?

jmenserg: I have already fixed these

Iryna-U (approved):

PR #63, opened 2021-08-12
clean and neat code first time

Evgen-Krasnenkov:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,19 @@
...
+            User iuser = iuserService.findByEmail(email);
+            if (iuser == null) {
+                return false;
+            } else {
+                if (email.equals(iuser.getEmail()) && password.equals(iuser.getPassword())) {
+                    return true;
+                }
+            }
+            return false;
+        }
```
Iryna-U, line 29: Please, don't complicate if-else constructions
```suggestion
        UserService userService = new UserService();
        User user = userService.findByEmail(email);
        if (user != null) {
            return user.getPassword().equals(password);
        } 
        return false;
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@
...
+        if (email == null) {
+            return null;
+        } else {
+            for (int posUsers = 0; posUsers < users.length; posUsers++) {
+                if (users[posUsers].getEmail().equals(email)) {
+                    return users[posUsers];
+                }
+            }
+            return null;
+        }
```
Iryna-U, line 27:
```suggestion
        }
        for (User user : users) {
            if (user.getEmail().equals(email)) {
                return user;
            }
        }
        return null;
```

Iryna-U (changes requested):

PR #71, opened 2021-08-14
Hw jv opp2

Evgen-Krasnenkov:
I've applied all the discussed changes. Thanks for reviewing my code.

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -13,19 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        if (email == null || password == null) {
-            return false;
-        } else {
-            UserService iuserService = new UserService();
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+
```
MaliukDaria, line 16: Do we need this empty line?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        if (email == null) {
+            return null;
```
MaliukDaria, line 19: Do we need this check? What will be the result of the expression `user.getEmail().equals(email)` if email == null?

MaliukDaria (changes requested):

PR #73, opened 2021-08-15
login and findByEmail methods were implemented.

pustoj0:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService service = new UserService();
+        User user = service.findByEmail(email);
+        if (user == null) {
+            return false;
+        } else {
```
MaliukDaria, line 20: Do we need `else` statement here?

MaliukDaria (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
pustoj0: Definetly not, I'll change

MaliukDaria (approved):

Max-Pochepets (approved):

PR #74, opened 2021-08-16
Hw jv oop3

Evgen-Krasnenkov:
Changed NULL check for email. Thanks for the review.

### DIFF
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -18,9 +18,9 @@ public class UserService {
         if (email == null) {
             return null;
         } else {
-            for (int posUsers = 0; posUsers < users.length; posUsers++) {
-                if (users[posUsers].getEmail().equals(email)) {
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@
...
+        if (email == null) {
+            return null;
+        } else {
+            for (User user : users) {
+                if (user.getEmail().equals(email)) {
+                    return user;
+                }
+            }
+            return null;
+        }
```
Max-Pochepets, line 27:
```suggestion
		if (email != null) {
	        for (User user : users) {
	            if (user.getEmail().equals(email)) {
	                return user;
	            }
	        }
	    }
        return null;
```
try not to complicate if-else construction

Max-Pochepets (changes requested):

Max-Pochepets (approved):

PR #75, opened 2021-09-03
Hw 1.2

v20stepanenko:
implemented task

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
@@ -11,6 +13,7 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = UserService.findByEmail(email);
+        return user.getPassword().equals(password);
     }
 }
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -2,6 +2,9 @@ package mate.academy.service;
 
 import mate.academy.model.User;
 
+import java.util.Arrays;
+
+
 public class UserService {
     private static final User[] users = new User[] {
             new User("bob@i.ua", "1234"),
@@ -14,7 +17,8 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
-        return null;
+    public static User findByEmail(String email) {
+        User user = Arrays.stream(users).filter(findUser -> findUser.getEmail().equals(email)).findFirst().orElse(null);
+        return user != null ? user : new User("none", "none");
     }
 }


src/main/java/mate/academy/service/UserService.java
```diff
@@ -1,20 +1,24 @@
 package mate.academy.service;
 
+import java.util.Arrays;
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
```
issugy, line 7: don't forget about whitespaces
```suggestion
    private static final User[] users = new User[] {
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -1,20 +1,24 @@
...
             new User("bob@i.ua", "1234"),
             new User("alice@i.ua", "1234")
     };
 
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
```
issugy, line 14: redundant empty line

src/main/java/mate/academy/service/UserService.java
```diff
@@ -1,20 +1,24 @@
...
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
-        return null;
+    public static User findByEmail(String email) {
```
issugy, line 19: lets try to avoid static, make method non-static please

src/main/java/mate/academy/service/UserService.java
```diff
@@ -1,20 +1,24 @@
...
+        User user = Arrays.stream(users).filter(findUser -> findUser.getEmail().equals(email))
+                .findFirst().orElse(null);
+        return user != null ? user : new User("none", "none");
```
issugy, line 22: It's kind of hardcoding, if user doesn't exist you should return `null` and not create a new one. If you already figure out what is `Optional`, you can return Optional.of(user), which your stream actually returns, or you can just return user, so it will be `null` if there is no user with such an email. Also, don't forget to check if user is `null` in `login` method

issugy (changes requested):
fix all from comments and place status 'r' in the doc

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email).isPresent()) {
+            User user = userService.findByEmail(email).get();
```
KseniiaMakarova, line 18: let's save this `userService.findByEmail(email)` into a local variable and reuse it

src/main/java/mate/academy/service/UserService.java
```diff
@@ -1,20 +1,25 @@
...
 
 public class UserService {
     private static final User[] users = new User[] {
             new User("bob@i.ua", "1234"),
             new User("alice@i.ua", "1234")
     };
-
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
```
KseniiaMakarova, line 14: what about this comment? it was not fixed

please remove this line

PR #79, opened 2021-09-09
implemented AuthenticationService and findByEmail methods

SarumanIsAlive:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        if (email.equals("bob@i.ua")) {
+            return users[0];
+        }
+        if (email.equals("alice@i.ua")) {
+            return users[1];
+        }
```
KseniiaMakarova, line 23: let's try to find the correct user in our `users` array using loop

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,12 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (email.equals("bob@i.ua") && password.equals("1234")) {
+            return true;
+        }
+        if (email.equals("alice@i.ua") && password.equals("1234")) {
+            return true;
+        }
```
KseniiaMakarova, line 19: can we use `findByEmail()` method of UserService to find the user by email?
and then compare that user's password with the given password

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #80, opened 2021-09-10
implemented login method

QQBratka:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,14 @@
...
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
+
```
KseniiaMakarova, line 6: redundant empty line
we don't need empty lines after class declarations

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,14 @@
...
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
```
KseniiaMakarova, line 5: let's not extend UserService

we can use `findByEmail()` if we create `new UserService()` object inside `login` method

KseniiaMakarova (changes requested):

PR #81, opened 2021-09-16
Implement findByEmail and login methods

AlexsisMoskovchuk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User finderUser = new UserService().findByEmail(email);
```
```suggestion
        User foundUser = new UserService().findByEmail(email);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,13 @@
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+
```
KseniiaMakarova, line 18: redundant empty line - we don't need empty lines after method signatures

PR #86, opened 2021-09-23

ShurionUA:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User userTest = UserService.findByEmail(email);
+        if (password.equals(userTest.getPassword())) {
+            return true;
+        }
         return false;
```
KseniiaMakarova, line 20: can we simplify this check?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User userTest = UserService.findByEmail(email);
+        if (password.equals(userTest.getPassword())) {
```
KseniiaMakarova, line 17: what if `userTest` is null?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
```
KseniiaMakarova, line 18: let's use `for-each` loop

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@
...
+            if (email == users[i].getEmail()) {
```
KseniiaMakarova, line 19: is it correct to compare strings like this?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```

src/main/java/mate/academy/service/AuthenticationService.java
ShurionUA: added null test

src/main/java/mate/academy/service/UserService.java
ShurionUA: I rewrote this loop

src/main/java/mate/academy/service/UserService.java
ShurionUA: No. I fixed all and pulled it again. Link in file changed.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User userTest = new UserService().findByEmail(email);
```
```suggestion
        User foundUser = new UserService().findByEmail(email);
```

KseniiaMakarova (approved):

PR #88, opened 2021-09-24

Nata-Vidzivashets:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) == null) {
```
KseniiaMakarova, line 15: let's save `userService.findByEmail(email)` into a local variable and reuse it

KseniiaMakarova (changes requested):

PR #93, opened 2021-10-01
I received the message described in the README file

nishchenkokateryna:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
...
 
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
```
KseniiaMakarova, line 19: we can use `for-each` loop here

PR #94, opened 2021-10-04
solution_01

Anatolii-Huryk:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for(User u : users){
```
```suggestion
        for(User user : users){
```
let's better not use shortened names for variables

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +12,9 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (userService.findByEmail(email) != null) {
+            return userService.findByEmail(email).getPassword().equals(password);
+        }
```
KseniiaMakarova, line 17: let's save `userService.findByEmail(email)` into a local variable

KseniiaMakarova (changes requested):

KseniiaMakarova (commented):
last step: run `mvn clean package` and fix checkstyle mistakes

PR #97, opened 2021-10-08
In AuthenticationService class login() method was implemented and in …

MykolaVoronov:
…UserService class findByEmail() method was implemented

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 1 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService service = new UserService();
+        if (service.findByEmail(email) != null
+                && service.findByEmail(email).getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
KseniiaMakarova, line 19: can we simplify this?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService service = new UserService();
+        if (service.findByEmail(email) != null
+                && service.findByEmail(email).getPassword().equals(password)) {
```
KseniiaMakarova, line 16: let's save `service.findByEmail(email)` into a variable

PR #99, opened 2021-10-12
implemented task

kateryna-deshchinska:
implemented task


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User userToLogin = UserService.findByEmail(email);
+        if (userToLogin != null && userToLogin.getPassword() == password) {
+        return true;
+        } else {
+            return false;
+        }
```
KseniiaMakarova, line 21: can we avoid this if-else?

pom.xml
```diff
@@ -56,5 +56,4 @@
             </plugins>
         </pluginManagement>
     </build>
-</project>
-
+</project>
```
KseniiaMakarova, line 59: we should add an empty line to the end of all files :)

KseniiaMakarova (approved):

PR #101, opened 2021-10-12
Implemented login and findByEmail methods

PineappleApplePinecone:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = new UserService().findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
issugy, line 20: Can be simplified to:
```suggestion
        return user != null && user.getPassword().equals(password);
```

PR #103, opened 2021-10-13
Implemented login method in AuthenticationService and findByEmail met…

Oleksii-kho:
…hod in UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User a = UserService.findByEmail(email);
+        if (a != null && email.equals(a.getEmail()) && password.equals(a.getPassword())) {
+            return true;
+        }
         return false;
     }
 }
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -14,7 +14,13 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+
+    public static User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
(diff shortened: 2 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User a = UserService.findByEmail(email);
```
KseniiaMakarova, line 16: is `a` an informative name?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,13 @@
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+
+    public static User findByEmail(String email) {
```
KseniiaMakarova, line 18: let's avoid using static

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,13 @@
...
+        for (int i = 0; i < users.length; i++) {
```
KseniiaMakarova, line 19: let's use `for-each` loop

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@
...
+        if (a != null && email.equals(a.getEmail()) && password.equals(a.getPassword())) {
```
KseniiaMakarova, line 17: we have already checked emails, right?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@
...
+            return true;
+        }
         return false;
```
KseniiaMakarova, line 20: can we simplify this part?

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,7 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = UserService.findByEmail(email);
```
KseniiaMakarova, line 16: does your code compile?
we should create an object of UserService, and call `findByEmail()` on this object

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #106, opened 2021-10-13
Login and email

MTWIB:
Login and findByEmail methods were implemented

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -11,6 +11,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        int i = 0;
+        while (i < 2) {
+            if (email.equals(UserService.users[i].getEmail())
+                    && password.equals(UserService.users[i].getPassword())) {
+                return true;
+            }
+            i++;
+        }
         return false;
     }
 }
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -11,13 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,14 @@
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,14 @@
...
+        int i = 0;
+        while (i < 2) {
```
KseniiaMakarova, line 19: can we use `for-each` loop?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != null) {
```
KseniiaMakarova, line 14: let's save this User who we get from `findByEmail()` method, into a local variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@
...
+            if (UserService.findByEmail(email).getPassword() == password) {
```
KseniiaMakarova, line 15: is it correct to compare strings with `==`?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@
...
+                return true;
```
KseniiaMakarova, line 16: try to remove this `if` - you don't really need it

KseniiaMakarova (changes requested):
the solution is good, but you need to run `mvn clean package` and fix the checkstyle mistakes

PR #107, opened 2021-10-13
Functions Login and findByEmail were implemented in a better way.

KostyaNikiforov:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,7 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        return user != null && user.getPassword() == password;
```
KseniiaMakarova, line 19: should we compare strings with `==`?

PR #110, opened 2021-10-13
login and findByEmail methods implementation with fixes

Savujkee:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User tmpUsr : users) {
```
```suggestion
        for (User user : users) {
```
it is difficult to read shortened names, let's not use them

PR #111, opened 2021-10-13
Jv oop lesson0

TankevychO:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,8 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
KseniiaMakarova, line 18:
```suggestion
        userService = new UserService(); // you have declared it at class level
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,8 @@
...
+        User user = userService.findByEmail(email);
+        return user != null && user.getPassword() == password;
```
KseniiaMakarova, line 20: is it correct to compare strings like this?

KseniiaMakarova (approved):

PR #113, opened 2021-10-13
Work2

Vitalbka:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+
```
KseniiaMakarova, line 16: we don't need empty lines after method signatures or class declarations

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+
```
KseniiaMakarova, line 18: same

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@
...
+        for (User users : users) {
```
KseniiaMakarova, line 19: does your code compile?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User users : users) {
```
KseniiaMakarova, line 18: what about this comment?

PR #114, opened 2021-10-13
Program correctly checks login and password 

George-Khomiak:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User userMail = new UserService().findByEmail(email);
```
```suggestion
        User foundUser = new UserService().findByEmail(email);
```
let's use more informative variable names

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User i : users) {
```
KseniiaMakarova, line 18: same here `i` or `u` is not informative

let's name it `user`

KseniiaMakarova (changes requested):

PR #115, opened 2021-10-14
implemented methods 'findByEmail' and 'login'

dima-barshay:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
KseniiaMakarova, line 17: please avoid using static

let's make this method non-static

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@
...
+        for (int i = 0; i < users.length; i++) {
```
KseniiaMakarova, line 18: let's better use `for-each` loop

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@
...
+            if (email.equals(users[i].getEmail()))w {
```
KseniiaMakarova, line 19: does your code compile?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != null
+                && email.equals(UserService.findByEmail(email).getEmail())
```
KseniiaMakarova, line 15: we have already checked emails, right?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@
...
+                && password.equals(UserService.findByEmail(email).getPassword())) {
```
KseniiaMakarova, line 16: let's save the User that we got from `findByEmail()` method, into a local variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@
...
+            return true;
+        }
         return false;
```
KseniiaMakarova, line 19: let's try to simplify this

PR #116, opened 2021-10-14
implemented login, find by email and find by password methods

Andrey-Sidorkin:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,11 @@
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        return (userService.findByEmail(email) != null
+                && userService.findByPassword(password) != null);
```
KseniiaMakarova, line 20: the logic is not quite correct here - we can have a User with matching email in our storage, and another User with matching password, and your code will give `true`

let's remove `findByPassword()` method and just compare the `password` that was passed to this method with the password of the User who was found by email

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
Andrey-Sidorkin: done!

PR #118, opened 2021-10-14
AuthenticationService and UserService for logging in completed

DanilOlgin:
Also changed java version from 11 to 8 in pom.xml in order to make project work.

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,16 +1,27 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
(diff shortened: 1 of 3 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,27 @@
...
-        return false;
+        User user = new UserService().findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        if (user.getPassword().equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
KseniiaMakarova, line 25:
```suggestion
        return user.getPassword().equals(password);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
...
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user : users) {
+            if (email == user.getEmail()) {
```
KseniiaMakarova, line 20: should we compare strings like this?

KseniiaMakarova (changes requested):

KseniiaMakarova (changes requested):
oops, your solution looks good, but we need to redo the task because you did the changes in `master` branch

you need to work in a separate branch

sorry that I didn't notice it before

PR #119, opened 2021-10-14
practice with OOP

DTyvoniuk:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
-        return null;
+    public static User findByEmail(String email) {
```
KseniiaMakarova, line 17: please avoid using static

let's make this method non-static

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@
...
+        for (int i = 0; i < users.length; i++) {
```
KseniiaMakarova, line 18: let's use `for-each` loop

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@
...
+            if (users[i].getEmail() == email) {
```
KseniiaMakarova, line 19: can we compare strings like this?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@
...
+                return users[i];
+            }
+        }
+        return new User(null,null);
```
KseniiaMakarova, line 23: let's just return null here

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,6 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+       return (UserService.findByEmail(email).getPassword() == password) ? true : false;
```
KseniiaMakarova, line 14: make sure that User is not null before you call `.getPassword()`

compare strings the right way

and you can get simplify the ternary operator here as well

PR #121, opened 2021-10-14
Two methods (login and findByEmail) were implemented

farfardesu:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+
```
KseniiaMakarova, line 16: we don't need empty lines after method signatures

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User users : users) {
```
KseniiaMakarova, line 18: does your code compile?

PR #124, opened 2021-10-15
Homework fixed

Andrey-Sidorkin:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,17 @@
...
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        return (userService.findByEmail(email) != null)
```
KseniiaMakarova, line 18: `userService.findByEmail(email)` let's save this into a local variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,17 @@
...
+                && (userService.findByEmail(email).getPassword() == password);
```
KseniiaMakarova, line 19: should we compare strings with `==`?

KseniiaMakarova (changes requested):

PR #125, opened 2021-10-15
Hw0

ZhukovVladyslav:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,13 @@
...
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
+
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null
+                && userService.findByPassword(password) != null) {
```
Nick97-git, line 8: Let's remove method findByPassword and compare the password with the field password of the user that will be found with the help of findByEmail method

ZhukovVladyslav: I think I fixed it

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,12 @@
...
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null
```
KseniiaMakarova, line 6: let's save `userService.findByEmail(email)` into a local variable and then reuse it

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,12 @@
...
+                && userService.findByEmail(email).getPassword() == password) {
```
KseniiaMakarova, line 7: is it correct to compare strings like this?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,12 @@
...
 
-    /**
-     * Find user by email. All users are stored in <code>private static final User[] users</code>
-     * @param email - the input parameter
-     * @return - user if his email is equal to passed email.
-     * Return <code>null</code> if there is no suitable user
-     */
     public User findByEmail(String email) {
+        for (User user : users) {
+            if (email == user.getEmail()) {
```
KseniiaMakarova, line 13: same here

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,12 @@
...
+            return true;
+        }
         return false;
```
KseniiaMakarova, line 10:
```suggestion
        return user != null && user.getPassword().equals(password);
```

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
ZhukovVladyslav: I'm sorry i did not fix this one

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,14 @@
...
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User ourUser = userService.findByEmail(email);
+        if (ourUser != null && ourUser.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
Nick97-git, line 12:
```suggestion
        return ourUser != null && ourUser.getPassword().equals(password);
```

PR #130, opened 2021-10-15
My hw

DTyvoniuk:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User i : users) {
```
KseniiaMakarova, line 18: is `i` an informative name?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@
...
+            if (i.getEmail() == email) {
```
KseniiaMakarova, line 19: is it ok to compare strings this way?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,15 @@
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) == null) {
```
KseniiaMakarova, line 15: save `userService.findByEmail(email)` into a local variable and then reuse it

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,15 @@
...
+            return false;
+        } else {
+            if (userService.findByEmail(email).getPassword() == password) {
+                return true;
+            } else {
+                return false;
+            }
```
KseniiaMakarova, line 22: can we simplify this?

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +12,14 @@
...
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        } else if(user.getPassword().equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
Nick97-git, line 23:
```suggestion
        return user != null && user.getPassword().equals(password))
```

DTyvoniuk: Спасибі)

PR #251, opened 2022-06-01

justpopova:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,28 @@
...
+
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User u : users) {
```
KseniiaMakarova, line 23:
```suggestion
        for (User user : users) {
```

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
```
KseniiaMakarova, line 5: why does it extend UserService? we don't need it)

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
justpopova: was thinking that it needs to be connected to each other? but will fix it now

KseniiaMakarova (approved):

PR #252, opened 2022-06-04
added solution

georgiy-nezbritskiy:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+       if (UserService.findByEmail(email) != null && UserService.findByEmail(email).getPassword().equals(password)) {
+           return true;
```
KseniiaMakarova, line 15: can we remove `if` and write `return a && b;`?

KseniiaMakarova (changes requested):
fix checkstyle - `mvn clean package`

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user;
+        user = new UserService().findByEmail(email);
```
KseniiaMakarova, line 17:
```suggestion
        User user = new UserService().findByEmail(email);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
...
+        return (user != null) && user.getPassword().equals(password);
```
KseniiaMakarova, line 18:
```suggestion
        return user != null && user.getPassword().equals(password);
```

KseniiaMakarova (approved):

PR #255, opened 2022-06-10
Hw 2 Lesson 0

verbanov:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService activeUserCheck = new UserService();
```
dSkarzh, line 16:
```suggestion
        UserService userService = new UserService();
```
this is not some check, this is object of `UserService` class, so this is service (hence the name), let's call it accordingly

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
...
+        User activeUser = new User(email, password);
```
dSkarzh, line 17: why do you need this? can you complete task without creating this object?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
...
+
+        if (activeUserCheck.findByEmail(email) != null) {
+            activeUser = activeUserCheck.findByEmail(email);
```
dSkarzh, line 20:
```suggestion
        if (userService.findByEmail(email) != null) {
```
this line means that user exists and user's email is the same as passed email, no need to check it once more

dSkarzh: you can also use & operator and simplify code even more

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User i : users) {
+            if (i.getEmail().equals(email)) {
+                return i;
+            }
+        }
```
dSkarzh, line 22:
```suggestion
        for (User user : users) {
            if (i.getEmail().equals(email)) {
                return i;
            }
        }
```
use understandable names

dSkarzh (changes requested):

verbanov:
deleted unused object, simplify solution

dSkarzh (changes requested):

PR #269, opened 2022-07-05
Hw oop

JK-devv:
first commit about methods 
second commit whit correct codeStyle

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -2,7 +2,7 @@ package mate.academy.service;
 
 import mate.academy.model.User;
 
-public class AuthenticationService extends UserService{
+public class AuthenticationService extends UserService {
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,8 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
```
dSkarzh, line 5: in order to use `findByEmail()` method you need to create `UserService` object, not extend from `UserService`

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,23 @@ public class UserService {
...
      */
+
     public User findByEmail(String email) {
+        for (User correctUser : users) {
+            if (correctUser.getEmail().equals(email)) {
+                return correctUser;
+            }
+        }
         return null;
+
```
dSkarzh, line 28:
```suggestion
        for (User user : users) {
            if (correctUser.getEmail().equals(email)) {
                return correctUser;
            }
        }
        return null;
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (User checkAllUsersCredit: getUsers()) {
+            if (checkAllUsersCredit.getEmail().equals(email)
+                    && checkAllUsersCredit.getPassword()
+                    .equals(password)) {
+                return true;
+            }
+        }
```
dSkarzh, line 22: it's the same logic as in `findByEmail()` but with password check
think how you can use logic from `findByEmail()` here

dSkarzh (changes requested):

KseniiaMakarova (changes requested):
- not all comments by Dmytro are fixed
- code style failed - run `mvn clean package` locally and fix mistakes

dSkarzh (approved):

PR #271, opened 2022-07-06
iimplemented login method, implemented findByEmail method

shevalye:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
...
     public boolean login(String email, String password) {
+
+        User user = UserService.findByEmail(email);
+        if (user != null) {
+
+            if (user.getPassword().equals(password)) {
+                return true;
+            }
+        }
+
```
dSkarzh, line 24:
```suggestion
        User user = UserService.findByEmail(email);
        if (user != null) {
            if (user.getPassword().equals(password)) {
                return true;
            }
        }
```
remove redundant empty lines in all project
think how you can use & operator to simplify your code

dSkarzh (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
dSkarzh, line 20:
```suggestion
        return user != null && user.getPassword().equals(password);
```

PR #274, opened 2022-07-08
upgrade code, implemented findByEmail method

LitvinOleksandr:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,13 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
-        return null;
+    public boolean findByEmail(String email, String password) {
```
dSkarzh, line 17: this method should return user, not boolean
please read carefully given comments and make changes to your code

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,13 @@ public class UserService {
...
+        boolean check = false;
+        for (int i = 0; i < users.length;i++) {
+            if ((users[i].getEmail().equals(email)) && (users[i].getPassword().equals(password))) {
+                check = true;
+            }
```
dSkarzh, line 22: use for each instead of for i, remove redundant parentheses, pay attention to spaces and codestyle

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,7 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
```
dSkarzh, line 13: this method should call `findByLogin()` and if user exists check if passwords match

dSkarzh (changes requested):

PR #283, opened 2022-07-13
All tests passed

max-bolotin:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
...
 
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
KseniiaMakarova, line 18: please don't use static methods

you can call this method from AuthenticationService if you create `new UserService()` object - and call this method on that object

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,22 @@
...
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = UserService.findByEmail(email);
+        if (user == null) {
+            return false;
+        } else return user.getPassword().equals(password);
```
KseniiaMakarova, line 20: code style check failed
run `mvn clean package` locally and fix mistake

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        } else {
+            return user.getPassword().equals(password);
+        }
```
KseniiaMakarova, line 23:
```suggestion
        if (user == null) {
            return false;
        } 
        return user.getPassword().equals(password);
```

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #285, opened 2022-07-13

SofiiaKazanivska:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User mail : users) {
```
dSkarzh, line 18: is this really mail? maybe user reflects better?

dSkarzh (changes requested):

dSkarzh (changes requested):

PR #287, opened 2022-07-17
Login and findByEmail methods were added

LubusUzh:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User mail : users) {
```
```suggestion
        for (User user : users) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService usr = new UserService();
```
```suggestion
        UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
...
+        User usrdata = new User(email, password);
```
KseniiaMakarova, line 17: why do we need to create this User?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
...
+        if (usr.findByEmail(email) != null) {
+            usrdata = usr.findByEmail(email);
```
KseniiaMakarova, line 19: let's save `userService.findByEmail(email)` into a local variable before `if` check

please use proper variable names
`usr`, `usrdata` are not good names:

- we should not shorten words
- we should use camelCase for variable names

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
LubusUzh: I don't need to create this User in that type.
I update my code in a new commit.

KseniiaMakarova (approved):

PR #290, opened 2022-07-17
"login" and "findByEmail" methods are integrated.

DioStesso1:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,12 @@ public class UserService {
...
     };
 
-    /**
-     * Find user by email. All users are stored in <code>private static final User[] users</code>
-     * @param email - the input parameter
-     * @return - user if his email is equal to passed email.
-     * Return <code>null</code> if there is no suitable user
-     */
     public User findByEmail(String email) {
+        for (User userUnit : users) {
```
dSkarzh, line 12:
```suggestion
        for (User user : users) {
```
let's call this variable user
try to use simple names when it's obvious what's happening

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,16 @@
...
-     */
     public boolean login(String email, String password) {
+        UserService userIsNotNull = new UserService();
+        User currentUser = new UserService().findByEmail(email);
+        if (userIsNotNull.findByEmail(email) != null
+                && userIsNotNull.findByEmail(email).getEmail().equals(email)
+                && currentUser.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
dSkarzh, line 14: let's improve naming and don't repeat method calls when we can use stored result
also don't create another object of user service when you can use existing one

dSkarzh (changes requested):

PR #294, opened 2022-07-19

JustElvis:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,8 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
```
dSkarzh, line 5: why do you need to extend from `UserService`?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
...
+        User user = new User(email, password);
+        if (findByEmail(email) == null) {
+            return false;
+        }
+        if (user.getEmail().equals(findByEmail(email).getEmail())
+                && user.getPassword() == findByEmail(email).getPassword()) {
+            return true;
+        } else {
+            return false;
+        }
```
dSkarzh, line 25: let's create `UserService` object, so we can call `findByEmail()` on that object
let's save result of `findByEmail()`, so we don't call method each time we want to use it's result
use && operator directly in return because && expression returns boolean

dSkarzh (changes requested):

KseniiaMakarova (approved):

PR #297, opened 2022-07-19
Methods findByEmail and login were implemented

ghost:
I made the findByEmail method static because the class includes an array of users and there is no need to create a separate object of this class to search for a specific user

### DIFF
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -2,6 +2,8 @@ package mate.academy.service;
 
 import mate.academy.model.User;
 
+import java.io.PushbackReader;
+
 public class UserService {
     private static final User[] users = new User[] {
             new User("bob@i.ua", "1234"),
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -2,6 +2,8 @@
 
 import mate.academy.model.User;
 
+import java.io.PushbackReader;
```
KseniiaMakarova, line 5: why do we need this import?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +16,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
KseniiaMakarova, line 19: please don't use static methods

you can call this method from AuthenticationService if you create `new UserService()` object - and call this method on that object

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/UserService.java
MyroslavVoskoboinyk: import java.io.PushbackReader was removed

src/main/java/mate/academy/service/UserService.java
MyroslavVoskoboinyk: static modifier has been removed and style mistakes were improved

KseniiaMakarova (approved):

PR #301, opened 2022-07-20
hw-OOP

Maksym-Korkosh:

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
...
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        } else {
+            return user.getPassword().equals(password);
+        }
```
```suggestion
        } else {
            return user.getPassword().equals(password);
        }
```
```suggestion
        } 
        return user.getPassword().equals(password);
```

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
KseniiaMakarova, line 22: we don't need else after return
```suggestion
        } 
        return user.getPassword().equals(password);
```

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #309, opened 2022-07-21
lesson 0 about OOP

UmyaFamilia:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService us = new UserService();
```
KseniiaMakarova, line 14:
```suggestion
        UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@ public class AuthenticationService {
...
+        if (us.findByEmail(email) != null) {
+            return us.findByEmail(email).getPassword().equals(password);
```
KseniiaMakarova, line 16: create a local variable for `userService.findByEmail()` and reuse it

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
```
KseniiaMakarova, line 18: better use for-each loop

KseniiaMakarova (changes requested):

PR #310, opened 2022-07-21
implemented login and findByEmail method

fliarkovskyi:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user != null) {
+            return user.getPassword() == password;
```
KseniiaMakarova, line 19: we should compare strings with equals()

KseniiaMakarova (changes requested):

Sofasmile (approved):

PR #316, opened 2022-07-29
Implemented methods "findByEmail" and "login"

Sachamunn:


liliia-ponomarenko (changes requested):
Let’s improve your solution!

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (User user : UserService.getUsers()) {
```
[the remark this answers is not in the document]
Sachamunn: Well done :)

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
```
ChumakAn, line 11: you don't need to create extra method and make it public, you can just refer to the users array as it's static

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
...
+
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
+        for (User user : getUsers()) {
```
ChumakAn, line 22: you can iterate through the users array without additional method

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
...
+    }
+
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
ChumakAn, line 21: don't make this method static

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+
+        if (user != null) {
+            return user.getEmail().equals(email) && user.getPassword().equals(password);
```
ChumakAn, line 19: we can simplify this, cause when you call findByEmail() method and it doesn’t return null, it means that user with this email already exists

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
```
ChumakAn, line 16: it's better to make an instance of UserService and then use the findByEmail() method

ChumakAn (changes requested):
Let's make some changes to make your solution

PR #318, opened 2022-08-06
Hw1

Mar4uk21:
Implemented task

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -4,15 +4,7 @@ import mate.academy.model.User;
 
 public class AuthenticationService {
 
-    /**
-     * Imagine that some user wants to login to your site.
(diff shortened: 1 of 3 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,19 @@
...
+
+
     public boolean login(String email, String password) {
+        for (User user : UserService.getUsers()) {
+            if (email.equals(user.getEmail())) {
+                if (password.equals(user.getPassword())) {
+                    return true;
+                }
+            }
+        }
```
liliia-ponomarenko, line 15: Try to avoid such code nesting. Let’s  use findByEmail() 😉

liliia-ponomarenko (changes requested):

liliia-ponomarenko (commented):
Let’s  improve  your solution  😎

liliia-ponomarenko (approved):

PR #320, opened 2022-08-09

AlinaOryshak:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,10 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for(User u:users){
+            if(u.getEmail()==email)
+                return u;
+        }
         return null;
```
liliia-ponomarenko, line 22: check code style. Don't use abbreviations.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService us=new UserService();
```
liliia-ponomarenko, line 16: the same here

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        if(us.findByEmail(email)!=null && us.findByEmail(email).getPassword()==password){
+            return true;
+        }else {
+            return false;
```
```suggestion
        return us.findByEmail(email) != null && us.findByEmail(email).getPassword() == password;
```
you can simplify  this code

liliia-ponomarenko (changes requested):

liliia-ponomarenko (approved):

PR #326, opened 2022-08-15
Hw 0 1

Matolch:
added two methods


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
-     * @param email - user's email
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (User user : UserService.users) {
+            if (email.equals(user.getEmail()) && password.equals(user.getPassword())) {
+                return true;
```
liliia-ponomarenko, line 18: Let’s use findByEmail() in this method 😉

PR #327, opened 2022-08-16
Implemented "login" and "findByEmail" methods

denys-domashevskyi:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService service = new UserService();
+
+        if (service.findByEmail(email) != null
+                && password.equals(service.findByEmail(email).getPassword())) {
```
liliia-ponomarenko, line 17: you can avoid code duplication if create a variable for  ```service.findByEmail(email)```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,13 @@ public class AuthenticationService {
...
+            return true;
+        } else {
+            return false;
```
liliia-ponomarenko, line 20: also, you can simplify this code using ```return <some_conditional>;```

liliia-ponomarenko (changes requested):
Let’s improve your solution 😉

liliia-ponomarenko (approved):
Great job! 👏

PR #332, opened 2022-08-18
implemented task/implemented login method/implemented findByEmail method

hpavlo:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        try {
```
liliia-ponomarenko, line 15: you don’t need to catch an exception in this case ;)

liliia-ponomarenko (approved):

PR #340, opened 2022-09-01
java-oop-lesson-0 / Solved the User Authentication Problem

YuriiHryniv:


YuriiHryniv:
Fixed checkstyle mistakes in the second version

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService us = new UserService();
+        for (User user : us.getUsers()) {
```
liliia-ponomarenko, line 18: why not use the findByEmail() method of the UserService?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@ public class UserService {
...
+        for (User user : users) {
+            if (email.equals(user.getEmail())) {
+                return user;
+            }
+        }
         return null;
     }
+
+    public User [] getUsers() {
+        return users;
```
liliia-ponomarenko, line 27: this method  will be unnecessary  when you fix the first comment

liliia-ponomarenko (changes requested):
Let’s  improve  your solution 😎

YuriiHryniv:
Thank you so much for your advice. I updated the solution. Don't mind the last 4th commit. I just tried to practice a bit :)

liliia-ponomarenko (approved):
good job 😎

PR #342, opened 2022-09-05
Add false

10pank007:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        return userService.findByEmail(email) != null
```
ChumakAn, line 15: it will be a good move to declare an separate variable for `userService.findByEmail(email)` as you use it twice.

10pank007: i made a new commit under. Declare new variable user

ChumakAn (approved):

PR #348, opened 2022-09-07
methods are implemented

igorgalenko:
have implemented a _login_ method in **AuthenticationService** and _findByEmail_ method in **UserService**

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

liliia-ponomarenko (approved):

PR #350, opened 2022-09-08

Andew-Miroshnikov:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = new UserService().findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
```suggestion
        return user != null && user.getPassword().equals(password);
```
you can simplify it

liliia-ponomarenko (approved):

PR #356, opened 2022-09-11

Danyl-Kniaziev:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+        if (user != null && user.getPassword() == password) {
+            return true;
+        }
         return false;
```
```suggestion
        return user != null && user.getPassword() == password;
```

liliia-ponomarenko (approved):

PR #358, opened 2022-09-13
the task was implemented

Roman-Usyk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
...
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        if (user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
liliia-ponomarenko, line 23: you can simplify  this code

liliia-ponomarenko (approved):

PR #361, opened 2022-09-15

nazariipertsak:


liliia-ponomarenko (changes requested):
Let’s  fix Java CI checks 😉

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService service = new UserService();
+        User userService = service.findByEmail(email);
+        if (userService != null && userService.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
ChumakAn, line 21: you can simplify this code
```suggestion
        return userService != null && userService.getPassword().equals(password);         
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService service = new UserService();
+        User userService = service.findByEmail(email);
```
ChumakAn, line 17: it's better to change naming for variables
```suggestion
        UserService userService = new UserService();
        User user = service.findByEmail(email);
```

PR #362, opened 2022-09-15
solved exercise "Working with services"

vazonchik:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        boolean result = false;
+        for (User user : UserService.users) {
```
liliia-ponomarenko, line 17: Let’s  use ```findByEmail()``` method 😉

liliia-ponomarenko (changes requested):
Let’s  improve your solution 😎

src/main/java/mate/academy/service/AuthenticationService.java
vazonchik: re-wrote by use "findByEmail()" method

liliia-ponomarenko (approved):

PR #364, opened 2022-09-16
Homework 0 

DeonOni:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,7 +3,7 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    static final User[] users = new User[] {
```
ChumakAn, line 6: let's better not make this array static if it's not mandatory

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.service;
 
+import static mate.academy.service.UserService.users;
```
ChumakAn, line 3: same for static here

ChumakAn (changes requested):

ChumakAn (changes requested):
Let's make your solution better, remember that we should use the static method only in case this method/ variable refers to a class, not the object. In order to get field values, we are using `objects` and their fields'  `getters`

ChumakAn (approved):

PR #365, opened 2022-09-16

apc10bbe:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userEmail = new UserService();
+        User user = userEmail.findByEmail(email);
+
```
ChumakAn, line 18: it's better to remove an empty line

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userEmail = new UserService();
```
ChumakAn, line 16: it's better to change naming 
```suggestion
        UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
...
+        User user = userEmail.findByEmail(email);
+
+        if (user != null) {
+            if (password.equals(user.getPassword())) {
+                return true;
+            }
+        }
         return false;
```
ChumakAn, line 24: let's simplify your code, try to avoid such nested constructions
```suggestion
	return user != null && password.equals(user.getPassword())
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
```
ChumakAn, line 18: what about replacing this for block with enhanced for?

ChumakAn (changes requested):
Good job, let's make your code better!

PR #368, opened 2022-09-18
F01

marchTooMuch:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,14 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
-        return null;
+    public static User findByEmail(String email) {
```
ChumakAn, line 17: It's better **not** to add `static` keyword to this method without a reason. What about creating removing static and regering to the UserService through the instance?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+        if (user == null) {
+            return false;
+        } else if (password == user.getPassword()) {
+            return true;
+        }
         return false;
```
ChumakAn, line 22: let's simplify this code 
```suggestion
        return user != null && password.equals(user.getPassword());
```
also, remember it's better to use equals() if you're comparing String objects

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,14 @@ public class UserService {
...
+
+        User result = null;
+        for (User user: users) {
+            if (email == user.getEmail()) {
+                result = user;
+            }
+        }
+        return result;
```
ChumakAn, line 25: let's simplify this one too and remove the redundant variable `result`

ChumakAn (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
+            if (email == user.getEmail()) {
```
ChumakAn, line 19: it's better to use equals method if we are comparing objects, like String in this case

ChumakAn (approved):
Well done! 👍

PR #370, opened 2022-09-19
Method "findByEmail" and "login" implemented

ashkil:
Static findByEmail method implemented in UserService. login method implemented in AuthenticationService


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
ChumakAn, line 17: it's better **not** to use `static` keyword, let's better use this method with UserService instance

ChumakAn (approved):

PR #372, opened 2022-09-19
Added methods users login and find users by email

romangolovatyi:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,6 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+public class AuthenticationService extends UserService {
```
KostyaNikiforov, line 3: In this case, it is better to use composition instead of inheritance

KostyaNikiforov (changes requested):
Let's try to simplify your solution)

src/main/java/mate/academy/service/AuthenticationService.java
romangolovatyi: Thank you for advice and revise.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private static final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        if (userService.findByEmail(email) != null) {
+            return userService.findByEmail(email).getPassword().equals(password);
+        }
         return false;
```
KostyaNikiforov, line 19: It is still can be simplified)
You can reuse the result of findByEmail method, so there is no need to call it twice

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private static final UserService userService = new UserService();
```
KostyaNikiforov, line 13:
```suggestion
    private final UserService userService = new UserService();
```

KostyaNikiforov (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
romangolovatyi: Sorry for the clarification.
I need to create a variable that finds and returns a user. And further, make manipulation with this variable. For example: 
```
User user =  userService.findByEmail(email);
        if (user != null) {
        }
        return false;
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
```
KostyaNikiforov, line 21: The second part of the expression will not be executed if the first part returns false ;)
```suggestion
        return user != null && user.getPassword().equals(password);
```

KostyaNikiforov (approved):

PR #378, opened 2022-09-22
Serhii jv-lesson-OOP

B1b1z9na:


KostyaNikiforov (changes requested):

KostyaNikiforov (commented):
It is a good solution:fire: But we can improve it a bit ;)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    UserService userService = new UserService();
```
KostyaNikiforov, line 15:
```suggestion
    private UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
...
+     
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
```
KostyaNikiforov, line 21:
```suggestion
         return user != null && user.getPassword().equals(password);
```

KostyaNikiforov (changes requested):
It still can be simplified)

KostyaNikiforov (commented):
Don't forget to run `mvn clean package` command before push.
Your code doest pass the tests)
Screenshot 2022-09-23 at 10 00 00

PR #380, opened 2022-09-23
made login and findByEmail methods

Alex-Kharchuk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
KostyaNikiforov, line 18: Let's make this var field, by moving it to the class level. So you will not need to create a new UserService object each time this method is called.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,13 @@ public class AuthenticationService {
...
+
+        if (userService.findByEmail(email) == null) {
+            return false;
+        }
+
+        User currentUser = userService.findByEmail(email);
```
KostyaNikiforov, line 24: Let's call the findByEmail method only once. And try to solve this task in one line

KostyaNikiforov (commented):
Let's improve your solution)

src/main/java/mate/academy/service/AuthenticationService.java
Alex-Kharchuk: Done in the new commit

PR #381, opened 2022-09-23
made login and findByEmail methods

Alex-Kharchuk:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
```
liliia-ponomarenko, line 18: you can use for-each in this case

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User currentUser = userService.findByEmail(email);
+        return userService.findByEmail(email) != null && currentUser.getPassword().equals(password) ? true : false;
```
```suggestion
        return userService.findByEmail(email) != null && currentUser.getPassword().equals(password);
```
you can simplify it

liliia-ponomarenko (approved):
good job 👍

liliia-ponomarenko (changes requested):
oh, you have problems with JavaCI checks, let’s fix them, and also you should fix my comments 😉

liliia-ponomarenko (approved):
LGTM!

PR #382, opened 2022-09-23
implemented task

alex-anofriev:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,7 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService use = new UserService();
```
hronograf, line 16: We can add field with type `UserService` to our class, so we won't need to create an instance every time
Also, consider another name for this variable (good one could be `userService`)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,7 @@ public class AuthenticationService {
...
+        return use.findByEmail(email) != null && (use.findByEmail(email).getPassword().equals(password));
```
hronograf, line 17: if `findByEmail` takes some time, this method will perform better if we introduce a variable for return object of `findByEmail` method

hronograf (changes requested):
Good!

hronograf:
Also it is better run `mvn clean package` before commit or push, so you will be confident that all tests passed

hronograf (approved):
Good job!

PR #383, opened 2022-09-25
implemented both method

Serhii-Prykhodko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,15 @@ public class AuthenticationService {
...
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+
+        if (user == null) {
+            return false;
+        }
+        return user.getPassword().equals(password);
```
liliia-ponomarenko, line 25:
```suggestion
        return user != null && user.getPassword().equals(password);
```
you can simplify  it 😉

liliia-ponomarenko (approved):
good job!

PR #392, opened 2022-11-14
implement findByEmail and login methods

Dmytro-Mk:


Nick97-git (changes requested):
run `mvn clean package` in terminal and fix code style errors
Screenshot 2022-11-23 at 15 21 57

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        return userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password);
```
Nick97-git, line 16: let's create a variable for the result of findByEmail, so we won't need to call it twice

Nick97-git (changes requested):

Nick97-git (approved):

PR #397, opened 2022-11-24
Users' email and password have been verified

Butovych:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,18 @@ public class AuthenticationService {
...
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password)) {
```
Nick97-git, line 17: let's create a variable for the result of findByEmail, so we won't need to call it twice

Nick97-git (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
...
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User byEmail = userService.findByEmail(email);
+        if (byEmail != null && byEmail.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
Nick97-git, line 22:
```suggestion
        User user = userService.findByEmail(email);
        return user != null && user.getPassword().equals(password);
```

Nick97-git (changes requested):

Nick97-git (approved):

PR #398, opened 2022-11-24
implemented task

Avassor3000:
Bugs are fixed

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -13,10 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        UserService x = new UserService();
-        User y = x.findByEmail(email);
-        if (y != null) {
-            if (y.getPassword().equals(password)) {
+        UserService user = new UserService();
+        User userEmail = user.findByEmail(email);
+        if (userEmail != null) {
+            if (userEmail.getPassword().equals(password)) {
                 return true;
             } else {
                 return false;
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -12,8 +12,9 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    UserService user = new UserService();
+    
     public boolean login(String email, String password) {
-        UserService user = new UserService();
         User userEmail = user.findByEmail(email);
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService x = new UserService();
```
KostyaNikiforov, line 16:
```suggestion
        UserService user = new UserService();
```
It is better to move this line to the class level. Otherwise, a new object of the type UserService will be created each time this method is called

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
...
+        User y = x.findByEmail(email);
```
KostyaNikiforov, line 17:
```suggestion
        User user = userService.findByEmail(email);
```
Do not use abbreviations;)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
...
+        if (y != null) {
+            if (y.getPassword().equals(password)) {
+                return true;
+            } else {
+                return false;
+            }
+        }
         return false;
```
KostyaNikiforov, line 25: Let's try shorting this code. You can do the same without any if blocks. Try to use a ternary operator where needed

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
...
      */
     public boolean login(String email, String password) {
+        UserService x = new UserService();
+        User y = x.findByEmail(email);
+        if (y != null) {
+            if (y.getPassword().equals(password)) {
+                return true;
+            } else {
+                return false;
+            }
```
KostyaNikiforov, line 23: For example
```suggestion
            return user.getPassword().equals(password));
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
+            if (user.getEmail().equals(email)) {
+                return user;
+            }
+        }
```
KostyaNikiforov, line 22: 🔥

KostyaNikiforov (changes requested):

PR #399, opened 2022-11-25
solution is ready

Mykola-Osolinskyi:

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -11,6 +11,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        var users = UserService.getUsers();
+        for (var i = 0; i < users.length; i++){
+            if (users[i].getEmail().equals(email) && users[i].getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
     }
 }
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -8,6 +8,10 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
+        return users;
+    }
+
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
      * @param email - the input parameter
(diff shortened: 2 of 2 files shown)


Nick97-git (changes requested):
run `mvn clean package` in terminal and fix code style errors
Screenshot 2022-11-25 at 14 15 17

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
```
Nick97-git, line 11: remove this method, use the method below for login method logic and don't use static, it's a bad practice

Nick97-git (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public User[] getUsers() {
```
Nick97-git, line 11: remove this method, use only findByEmail

Nick97-git (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService service = new UserService();
+        User user = service.findByEmail(email);
+        return (user != null && user.getPassword().equals(password));
```
KostyaNikiforov, line 18:
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0;i < users.length; i++) {
+            if (users[i].getEmail().equals(email)) {
```
KostyaNikiforov, line 19: Let's use a for-each loop. We don't need an index here

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #403, opened 2022-11-27
Implemented findByEmail method

OleksiiZaloznyi:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
...
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
KostyaNikiforov, line 22:
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (users[i].getEmail().equals(email)) {
+                return users[i];
+            }
+        }
```
KostyaNikiforov, line 22: Let's use for-each loop, we don't need an index here

KostyaNikiforov (changes requested):
Good solution!:fire:

KostyaNikiforov (approved):

PR #404, opened 2022-11-27
implemented login method

imhatep:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +11,9 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        return userService.findByEmail(email).getEmail().equals(email)
+               && userService.findByEmail(email).getPassword().equals(password);
```
boroda4436, line 17: Better call method findByEmail only once, not twice:
```suggestion
	User user = userService.findByEmail(email);
        return user.getEmail().equals(email) && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -7,14 +7,20 @@ public class UserService {
...
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
+
     public User findByEmail(String email) {
-        return null;
+        int userMail = 0;
+        for (int i = 0; i < users.length; i++) {
+            if (users[i].getEmail().equals(email)) {
+                userMail = i;
```
boroda4436, line 21: better call return here directly

src/main/java/mate/academy/service/UserService.java
```diff
@@ -7,14 +7,20 @@ public class UserService {
...
+            }
+        }
+        return users[userMail];
```
boroda4436, line 24: If nothing was found - you should return null
```suggestion
        return null;
```

boroda4436 (changes requested):

boroda4436 (approved):

PR #406, opened 2022-11-28
Login and Authentication Services were added

Andrew-Yakhimets:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+
```
KostyaNikiforov, line 18: Remove empty line

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+
```
KostyaNikiforov, line 16: the same

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
...
+        UserService userService = new UserService();
```
KostyaNikiforov, line 17: It is better to move this line to the class level. Becouse otherwise a new object of the type UserService will be created each time this method is called

KostyaNikiforov (changes requested):

KostyaNikiforov (changes requested):
Please, run the `mvn clean package` command one more time;)

KostyaNikiforov (approved):

PR #409, opened 2022-11-29
login and findByEmail methods implemented

taras-zhyravetskyi:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+
+        UserService userService = new UserService();
```
KostyaNikiforov, line 17: It is better to move this line to the class level because otherwise a new object of a type UserService will be created each time this method is called.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+
```
KostyaNikiforov, line 16: Remove empty line

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+
+        return user == null ? false : password.equals(user.getPassword());
```
KostyaNikiforov, line 20:
```suggestion
        return user != null && password.equals(user.getPassword());
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -7,14 +7,19 @@ public class UserService {
...
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
+
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (email.equals(users[i].getEmail())) {
+                return users[i];
+            }
+        }
```
KostyaNikiforov, line 22: Let's use a for-each loop. We don't need an index

KostyaNikiforov (changes requested):

KostyaNikiforov (approved):

PR #414, opened 2022-11-30
added solution for the task

Dastya:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        UserService service = new UserService();
```
KostyaNikiforov, line 17: It is better to move this line to the class level. Becouse otherwise a new object of the type UserService will be created each time this method is called

KostyaNikiforov (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
Dastya: done

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,26 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
             new User("bob@i.ua", "1234"),
             new User("alice@i.ua", "1234")
     };
+    long abc = 1000L;
+    float bca = 1000f;
```
Nick97-git, line 11: do you use these variables somewhere?

Nick97-git (changes requested):

Nick97-git (approved):

PR #415, opened 2022-12-01
all tests passed, checkstyle passed too

Maks4u:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@ public class UserService {
...
+            if (users[i].getEmail().equals(email)) {
+                return users[i];
+            }
+        }
         return null;
     }
+
+    public static User[] getUsers() {
+        return users;
+    }
```
KostyaNikiforov, line 28: Remove this method, let's use only findByEmail

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +12,15 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService test = new UserService();
```
KostyaNikiforov, line 15:
```suggestion
        UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +12,15 @@ public class AuthenticationService {
...
+
+        for (int i = 0; i < UserService.getUsers().length; i++) {
```
KostyaNikiforov, line 17: Let's try to solve this method without a loop

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +12,15 @@ public class AuthenticationService {
...
+            if (test.findByEmail(email) != null
```
KostyaNikiforov, line 18: Also, you can do it without any if-block ;)

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
```
KostyaNikiforov, line 18: Let's use for-each method. We don't need an index here

KostyaNikiforov (changes requested):

Nick97-git (changes requested):

PR #416, opened 2022-12-01
login and findByEmail was edded

Pandoryms:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@ public class UserService {
...
+            if (users[i].getEmail().equals(email)) {
+                return users[i];
+            }
+        }
         return null;
     }
+
+    public static User[] getUsers() {
+        return users;
+    }
```
KostyaNikiforov, line 28: Remove this method, let's use only findByEmail

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,15 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService authentication = new UserService();
```
KostyaNikiforov, line 14:
```suggestion
        UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,15 @@ public class AuthenticationService {
...
+
+        for (int i = 0; i < UserService.getUsers().length; i++) {
```
KostyaNikiforov, line 16: Let's try to solve this method without a loop

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,15 @@ public class AuthenticationService {
...
+            if (authentication.findByEmail(email) != null
```
KostyaNikiforov, line 17: Also, you can do it without any if-block ;)

KostyaNikiforov (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
```
KostyaNikiforov, line 18: Let's use for-each method. We don't need an index here

KostyaNikiforov (changes requested):

PR #423, opened 2022-12-15
Implemented login and fyndByEmail methods

Yehor-Cherkashyn:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User result = new UserService().findByEmail(email);
```
boroda4436, line 16: Please, create a separate variable of type UserService in the AuthenticationService class (for example at line 6) and use it in this line.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
...
+        if (result == null) {
+            return false;
+        }
+        String em = result.getEmail();
+        String pass = result.getPassword();
```
boroda4436, line 21: Better use more meaningful naming:
```suggestion
        String userEmail = result.getEmail();
        String userPassword = result.getPassword();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User result = new UserService().findByEmail(email);
+        if (result == null) {
```
boroda4436, line 17: Better use more meaningful naming instead of `result`. For example you could use `user` variable name

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
...
+            return false;
+        }
+        String em = result.getEmail();
+        String pass = result.getPassword();
+        if (em.equals(email) && pass.equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
boroda4436, line 26: You can simplify your code:
```suggestion
        return em.equals(email) && pass.equals(password);
```

boroda4436 (changes requested):

boroda4436 (approved):

PR #427, opened 2023-01-05
method login and method findByEmail were implemented

oleksiistepaniak:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
```suggestion
        return user != null && user.getPassword().equals(password);
```
You can simplify it

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
```
liliia-ponomarenko, line 18:
```suggestion
        for (User user : users) {
```

liliia-ponomarenko (approved):

src/main/java/mate/academy/service/AuthenticationService.java
oleksiistepaniak: How can I implement the suggested changes? I've opened this project in the environment of development on my desktop and then I've implemented the suggested changes. After that I've created a new commit. Do I need to create a new pull request or what? Thank you in advance.

PR #429, opened 2023-01-16
Login is fixed

IvanZuravlev228:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User userX = userService.findByEmail(email);
```
liliia-ponomarenko, line 17:
```suggestion
        User user = userService.findByEmail(email);
```

liliia-ponomarenko (approved):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        if (userX != null && userX.getPassword().equals(password)){
+            return true;
+        }
         return false;
```
liliia-ponomarenko, line 21: Also, you can simplify  it 
```suggestion
        return user != null && userX.getPassword().equals(password);
```

liliia-ponomarenko (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (email.equals(users[i].getEmail())) {
+                return users[i];
+            }
+        }
```
KseniiaMakarova, line 22: it is more elegant to use for-each loop

KseniiaMakarova (approved):

PR #433, opened 2023-01-22
Mate HW lesson-0

KosMarch:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,14 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        if (email == null) {
+            return null;
+        }
```
liliia-ponomarenko, line 20: Is this check really necessary? What will be returned if email = null without this check?

liliia-ponomarenko (approved):

PR #436, opened 2023-01-24
1

ElDaunito:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService user = new UserService();
+        return user.findByEmail(email) != null
+                && user.findByEmail(email).getPassword().equals(password);
```
boroda4436, line 16: It is better to call method only once, not twice
```suggestion
	User user = userService.findByEmail(email)
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService user = new UserService();
```
boroda4436, line 14: Please use proper variable naming
```suggestion
        UserService userService = new UserService();
```

boroda4436 (changes requested):

PR #438, opened 2023-01-24
implemented task

sanya-1xx8:


sanya-1xx8:
complit

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+
+        if (userService.findByEmail(email) != null) {
+            User user = userService.findByEmail(email);
```
boroda4436, line 19: Better call method `findByEmail` only once, not twice:
```suggestion
	User user = userService.findByEmail(email);
        if (user != null) {
```

boroda4436 (changes requested):

boroda4436 (approved):

PR #441, opened 2023-01-25
implemented authentication methods

denys-mg:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+
+        if (user == null) {
+            return false;
+        }
+
+        if (user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
boroda4436, line 26: You can simplify your code:
```suggestion
        return user.getPassword().equals(password);
```

boroda4436 (approved):

PR #443, opened 2023-01-26
methods findByEmail, login was updated 

potrap:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,11 @@
...
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
+
     public boolean login(String email, String password) {
-        return false;
+        if (new UserService().findByEmail(email) == null) {
+            return false;
+        }
+        return new UserService().findByEmail(email).getPassword().equals(password);
```
liliia-ponomarenko, line 9: Can you create only one instance of UserService? Can you call the method `findByEmail(email)` only once?

liliia-ponomarenko (changes requested):

liliia-ponomarenko (approved):

PR #445, opened 2023-01-27

OlehSarapuk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        if (UserService.findByEmail(email) != null
+                && UserService.findByEmail(email).getPassword().equals(password)) {
```
liliia-ponomarenko, line 15: Can you call the method `findByEmail(email)` only once?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
liliia-ponomarenko, line 17: Why is this method static? Don't change the method signature

liliia-ponomarenko (changes requested):

liliia-ponomarenko (approved):

PR #449, opened 2023-01-28

OlStepp:

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null) {
+            if (userService.findByEmail(email).getPassword().equals(password)) {
```
liliia-ponomarenko, line 18: Can you call the method findByEmail(email) only once?
You also can avoid conditions nested

liliia-ponomarenko (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User someone = userService.findByEmail(email);
+        if (someone != null && someone.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
```suggestion
        return someone != null && someone.getPassword().equals(password);
```
you can simplify it

liliia-ponomarenko (approved):

PR #455, opened 2023-01-30
implementation of auth method

kromch:


boroda4436 (changes requested):
Please read the task description more carefully.

PR #457, opened 2023-01-30
Home-work-jv-oop-lesson-0

V-Vasylchuk:
implement a login and findByEmail method

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
...
     public boolean login(String email, String password) {
-        return false;
+        User user = new UserService().findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        String userEmail = user.getEmail();
+        String userPassword = user.getPassword();
+
+        return userEmail.equals(email) && userPassword.equals(password);
```
liliia-ponomarenko, line 23:
```suggestion
        return userPassword.equals(password);
```
you checked it in `findByEmail` method

liliia-ponomarenko (approved):

PR #468, opened 2023-01-31

pechivo:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        if (email == null && password == null) {
+            return false;
+        }
+        return findByEmail(email) != null && password.equals(findByEmail(email).getPassword());
```
boroda4436, line 17: You should call the method `findByEmail` only once, not twice

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,6 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+public class AuthenticationService extends UserService {
```
boroda4436, line 3: This is very bad practice. Please do not inherit services. We will talk about it next week

boroda4436 (changes requested):

boroda4436 (approved):

PR #470, opened 2023-01-31
Added login method in AuthenticationService class, Added findByEmail …

De1eF:
…method in UserService class

### DIFF
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -14,7 +14,15 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
-        return null;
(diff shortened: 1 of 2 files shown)


boroda4436 (changes requested):
Java CI check has not passed. Please fix your checkstyle issues

boroda4436 (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
petro-ushchuk, line 17: This method can not be static.

petro-ushchuk (changes requested):

PR #472, opened 2023-02-01
make method find_by_email and login method

yurihhhT:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User u : users) {
```
liliia-ponomarenko, line 18:
```suggestion
        for (User user : users) {
```
Don’t  use abbreviations  for variable names

liliia-ponomarenko (approved):

PR #475, opened 2023-02-01

outref:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User[] users = UserService.getUsers();
+        for (User user : users) {
+            if (user.getEmail().equals(email) && user.getPassword().equals(password)) {
+                return true;
```
liliia-ponomarenko, line 19: Let's use findByEmail(String email) method for avoiding  code duplicate

liliia-ponomarenko (changes requested):

boroda4436 (approved):

PR #483, opened 2023-02-02
implemented UserService and AuthService

Skrudj:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword() == password) {
+            return true;
+        }
         return false;
```
boroda4436, line 21: You can simplify your code:
```suggestion
        return user != null && user.getPassword() == password;
```

boroda4436 (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        String s1 = "412341324";
```
sarakhmen, line 16: This redundant variable should be removed

sarakhmen (approved):

PR #487, opened 2023-02-03
solve

AnonimToster:
make findByEmail method static and implements its realization. implementation login method

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import  mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
...
+        User founded = UserService.findByEmail(email);
+        if (founded != null) {
+            if (founded.getPassword() == password) {
+                return true;
+            } else {
+                return false;
+            }
+        } else {
+            return false;
+        }
```
boroda4436, line 25: Let's simplify the code and write it in a single line

boroda4436 (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
boroda4436: try to use only one return statement in your solution

AnonimToster: Now the Login method uses only one return keyword

PR #490, opened 2023-02-05
fix AuthenticationService.login & UserService.findByEmail methods

NickolayVoloshchuk:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
karma-o, line 17: Let's not change the signature of the method, it should be non-static, and we should create a new object of that `UserService` in the `AuthenticationService` to be able to call this method

karma-o (changes requested):

PR #492, opened 2023-02-06
check user autentification

yuriy49:
in class userservice make comparing email
in autentification check email by using UserService methon and check user object password

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -1,8 +1,8 @@
 package mate.academy.model;
 
-public class User {
-    private String email;
-    private String password;
+public class User{
```
petro-ushchuk, line 3: Java Cl checks failed. Before making a Pull Request (or a new commit to an already existing PR), you need to write the mvn clean package command in the terminal and make sure that all checks pass. After that, the mentor will be able to review the homework

petro-ushchuk (changes requested):

PR #499, opened 2023-03-30

eugenetolkunov:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null) {
+            if (password.equals(userService.findByEmail(email).getPassword())) return true;
+        }
```
sarakhmen, line 19: I'd recommend calling findByEmail() method only once. Also, these ifs could be easily simplified and the solution will contain only one line (consider using '&&' operator)

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User i : users) {
```
sarakhmen, line 18: let's make your variable more readable
```suggestion
        for (User user: users) {
```

sarakhmen (changes requested):
Let's improve your solution ;)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +12,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email, password) != null) {
+            return true;
+        }
         return false;
```
sarakhmen, line 19:
```suggestion
        return userService.findByEmail(email, password) != null
```
let's remove redundant if block

sarakhmen (approved):
Good job ;)

PR #500, opened 2023-03-30
added email and password checking

viprko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +12,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        return userService.findByEmail(email) != null
+                && password.equals(userService.findByEmail(email).getPassword());
```
sarakhmen, line 17: I'd recommend calling findByEmail() method only once

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User i : users) {
```
sarakhmen, line 18:
```suggestion
        for (User user : users) {
```
I'd also recommend making this name more informative

sarakhmen (approved):

PR #503, opened 2023-04-01
implemented task requirements

Tupadzaky:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
         return false;
```
sarakhmen, line 22:
```suggestion
        user != null && user.getPassword().equals(password);
```
Let's simplify this construction :)

sarakhmen (approved):

PR #505, opened 2023-04-02
Hw 3

ASnazar:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,18 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+    private User user;
```
sarakhmen, line 16: It's better to remove this class field and create local variable for user directly in the method

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
```
sarakhmen, line 6: These comments should be placed directly above your login method as they are documenting it

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,18 @@ public class AuthenticationService {
...
     public boolean login(String email, String password) {
+        user = userService.findByEmail(email);
+
+        if (user == null) {
+            return false;
+        } else if (user.getPassword() == password && user.getEmail() == email) {
+            return true;
+        }
+
         return false;
```
sarakhmen, line 27: Let's remove redundant if-else:
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,26 @@ public class UserService {
...
+
+
+
+
+
+
+
+
+
+
```
sarakhmen, line 40: Please also remove these redundant empty lines

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,26 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User n : users) {
```
sarakhmen, line 18: I'd recommend renaming this local variable

sarakhmen (changes requested):
Good job, but let's improve your solution :)

PR #507, opened 2023-04-03

pkobevko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+
+        return user == null ? false : user.getPassword().equals(password);
```
sarakhmen, line 19: let's make your code more readable:
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (users[i].getEmail().equals(email)) {
+                return users[i];
+            }
+        }
```
sarakhmen, line 22: Here you can use foreach loop instead

sarakhmen (approved):

PR #513, opened 2023-04-07
hw-02 (jv-oop) from lesson-0

nechupara:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
...
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+
+        if (user != null && user.getPassword() == password) {
+            return true;
+        }
+
         return false;
```
sarakhmen, line 24: Let's remove this redundant 'if' construction and also use equals() method to compare passwords:
```suggestion
        return user != null && user.getPassword.equals(password)
```

sarakhmen (changes requested):
Good job, but let's improve your solution ;)

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
...
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user : users) {
+            if (user.getEmail() == email) {
```
sarakhmen, line 20: let's also use equals() here

src/main/java/mate/academy/service/AuthenticationService.java
nechupara: 'if' construction has been changed to a `return user != null && user.getPassword().equals(password)` and also `user.getEmail() == email` in 'if' statement been changed to `user.getEmail().equals(email)` in UserService file.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,22 @@
...
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+
```
sarakhmen, line 19: let's remove redundant empty lines next time ;)

sarakhmen (approved):
Good job :)

PR #514, opened 2023-04-07

Valdemar619:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email, password) != null) {
+            return true;
+        }
         return false;
```
sarakhmen, line 18: we can simplify the code this way:
```suggestion
        return userService.findByEmail(email, password) != null;
```

sarakhmen (approved):
Good job :)

PR #515, opened 2023-04-08
implemented methods

MykytaKat:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userServise = new UserService();
+        User user = userServise.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
sarakhmen, line 21: Let's simplify your code and avoid using 'if' here
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user : users) {
+            if (user.getEmail() == email) {
```
sarakhmen, line 19: let's compare emails using equals() method

sarakhmen (changes requested):

PR #526, opened 2023-04-12
Hw2

Fafakepa:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
sarakhmen, line 17: let's not make this method a static one

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
...
+        if (email.equals(users[0].getEmail())) {
+            return users[0];
+        } else if (email.equals(users[1].getEmail())) {
+            return users[1];
+        }
```
sarakhmen, line 22: what if your users array contains more than 2 users in it? I'd recommend using loop here

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User test = UserService.findByEmail(email);
```
sarakhmen, line 16:
```suggestion
        User user = UserService.findByEmail(email);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
...
+        if (test != null) {
+            if (password.equals(test.getPassword())) {
+                return true;
+            }
+        }
         return false;
```
sarakhmen, line 22: Let's simplify it this way:
```suggestion
        return user != null && user.getPassword().equals(password);
```

sarakhmen (changes requested):

sarakhmen (approved):

PR #527, opened 2023-04-12

m-pnsk:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
maxline, line 17: please do not change the method modifiers. fix the solution without using `static` for this method

maxline (changes requested):
good job! let's improve the solution

PR #528, opened 2023-04-13
test passed

MykNich:
methods are working and test passed for all cases

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,6 +1,7 @@
 package mate.academy.service;
 
 public class AuthenticationService {
+
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
@@ -11,6 +12,7 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        return (email == "bob@i.ua" && password == "1234")
+                || (email == "alice@i.ua" && password == "1234");
     }
 }
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -15,6 +15,13 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        if (email == "bob@i.ua") {
+            return users[0];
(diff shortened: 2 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,13 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        if (email == "bob@i.ua") {
```
maxline, line 18: you should use vise verse algorithm
use loop for iterating users
` for (int i = 0; i < users.length; i++) {`

and inside this loop check which `user` has `email`  that equals `email` from method input parameter

maxline (changes requested):
let's improve the solution

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +14,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        return user.getPassword() == password;
```
sarakhmen, line 22: I'd recommend simplifying it this way:
```suggestion
        return user != null && user.getPassword().equals(password);
```
Do not forget to use equals() method instead of == operator

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (email == users[i].getEmail()) {
+                return users[i];
+            }
+        }
         return null;
+
```
sarakhmen, line 24: redundant empty line

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (email == users[i].getEmail()) {
```
sarakhmen, line 19: let's compare emails using equals() operator

sarakhmen (changes requested):

PR #531, opened 2023-04-13
implemented task/implemented login method/implemented findByEmail method

mpankrad:
implemented task/implemented login method/implemented findByEmail method

### DIFF
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -1,7 +1,7 @@
 package mate.academy.model;
 
 public class User {
-    private static String password;
+    private String password;
     private String email;
 
     public User(String email, String password) {
@@ -13,7 +13,7 @@ public class User {
         return email;
     }
 
-    public static String getPassword() {
+    public String getPassword() {
         return password;
     }
 }
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -2,7 +2,7 @@ package mate.academy.model;
 
 public class User {
     private String email;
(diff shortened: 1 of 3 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -1,8 +1,8 @@
 package mate.academy.model;
 
 public class User {
+    private static String password;
```
sarakhmen, line 4: Why did you make the password a static field? Let's revert this change

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
sarakhmen, line 17: let's also make this method non-static

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+        if (user != null && User.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
sarakhmen, line 20: could be simplified by this way:
```suggestion
        return user != null && user.getPassword().equals(password);
```

sarakhmen (changes requested):

sarakhmen (approved):

PR #536, opened 2023-04-18
implement a login method and findByEmail method

maksigors:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+
```
sarakhmen, line 16: this empty line is redundant, let's remove it

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
...
+        UserService us = new UserService();
+        User user = us.findByEmail(email);
+        if (user != null) {
+            if (password.equals(user.getPassword())) {
+                return true;
+            }
+        }
+
```
sarakhmen, line 24: the same here

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+
+        UserService us = new UserService();
```
sarakhmen, line 17: let's make your naming more clear:
```suggestion
        UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
...
+        User user = us.findByEmail(email);
+        if (user != null) {
+            if (password.equals(user.getPassword())) {
+                return true;
+            }
+        }
```
sarakhmen, line 23: this block of code could be easily simplified this way:
```suggestion
    return user != null && user.getPassword().equals(password);
```

sarakhmen (changes requested):

PR #545, opened 2023-05-22
implemented task

irynamekh:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,10 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+
```
nacenik, line 6: delete empty line 
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user.getPassword().equals(password) && user != null) {
```
nacenik, line 19: What happens if user will be `null`? Could we call any method from `null`? 
You should check first on `null` and after checking password equality
```suggestion
        if (user != null && user.getPassword().equals(password)) {
```

nacenik (changes requested):
Good solution, let's improve it :) 

Don't forget to use the command  `man clean compile` on your local machine to check and fix your check style bugs

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +17,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
```
KseniiaMakarova, line 20: let's better use for-each loop

KseniiaMakarova (approved):

PR #553, opened 2023-06-01
implemented login and findbyemail methods

Pafmortim:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,13 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
```
KseniiaMakarova, line 18: better use for-each loop here

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
+
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User i : users) {
```
nacenik, line 18: `i` is a bad name for variable 
```suggestion
        for (User user : users) {
```

nacenik (changes requested):
Let's fix variable name

PR #554, opened 2023-06-03
added solution

DaniilNepelev:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User userIndex = userService.findByEmail(email);
+        boolean checkThePassword = userService.isPasswordCorrect(email, password);
+
+        return userIndex != null && checkThePassword;
```
nacenik, line 20:
```suggestion
        return userIndex != null && user.getPassword().equals(password);
```

nacenik (changes requested):
Good solution :) Let's improve it.

PR #556, opened 2023-06-04
implemented task

OleksandrSaven:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@ public class UserService {
...
+            if (user.getEmail().equals(email)) {
+                return user;
+            }
+        }
         return null;
     }
+
+    public static User[] getUsers() {
+        return users;
+    }
```
KseniiaMakarova, line 28:
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User[] users = UserService.getUsers();
```
KseniiaMakarova, line 16: you should create an object of userService class and call method findByEmail() on this object

then compare passwords

KseniiaMakarova (changes requested):

nacenik (approved):
Good job 👍

PR #558, opened 2023-06-06
HW / jv-oop-lesson-0-solution

keda-vlad:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,15 @@ public class AuthenticationService {
...
+    private static final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User potentialUser = userService.findByEmail(email);
+        if (potentialUser == null) {
+            return false;
+        } else {
+            return potentialUser.getPassword().equals(password);
+        }
```
KseniiaMakarova, line 24:
```suggestion
        } 
        return potentialUser.getPassword().equals(password);
```

KseniiaMakarova (changes requested):

PR #561, opened 2023-06-10
Hw oop 0

CodeWebWeaver:
Implemented requested methods make it possible to find user by email and check successful login or not.
 Code passed through mvn clean package

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
@@ -11,14 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String emailEntered, String passwordEntered) {
-        if (UserService.findByEmail(emailEntered) == null) {
+        UserService userService = new UserService();
+        User findedUser = userService.findByEmail(emailEntered);
+        if (findedUser == null) {
             return false;
         } else {
-            if (passwordEntered.equals(UserService.findByEmail(emailEntered).getPassword())) {
-                return true;
-            } else {
-                return false;
-            }
+            return passwordEntered.equals(findedUser.getPassword());
         }
     }
 }
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -16,9 +16,7 @@ public class AuthenticationService {
         if(UserService.findByEmail(emailEntered) == null){
             return false;
         } else {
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
KseniiaMakarova, line 17:
```suggestion
    public User findByEmail(String email) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,20 @@ public class AuthenticationService {
...
-     * @param password - user's password
+     * @param emailEntered - user's email
+     * @param passwordEntered - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
-    public boolean login(String email, String password) {
-        return false;
+    public boolean login(String emailEntered, String passwordEntered) {
+        if (UserService.findByEmail(emailEntered) == null) {
```
KseniiaMakarova, line 14: create an object and call this method on an object

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,20 @@ public class AuthenticationService {
...
+            return false;
+        } else {
+            if (passwordEntered.equals(UserService.findByEmail(emailEntered).getPassword())) {
+                return true;
+            } else {
+                return false;
+            }
+        }
```
```suggestion
            return false;
        } 
        return passwordEntered.equals(user.getPassword());
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
...
+        for (User oneInAllUsers : users) {
```
```suggestion
        for (User user : users) {
```

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
...
-    public boolean login(String email, String password) {
-        return false;
+    public boolean login(String emailEntered, String passwordEntered) {
+        UserService userService = new UserService();
+        User findedUser = userService.findByEmail(emailEntered);
+        if (findedUser == null) {
+            return false;
+        } else {
+            return passwordEntered.equals(findedUser.getPassword());
+        }
```
```suggestion
        } 
        return passwordEntered.equals(findedUser.getPassword());
```

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
Skarbey: Thanks for advice. I fixed it

PR #565, opened 2023-06-13
implemented task

OlegStashkiv:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
KseniiaMakarova, line 17:
```suggestion
    public User findByEmail(String email) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = UserService.findByEmail(email);
```
KseniiaMakarova, line 16: create an object and call the method on an object

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #566, opened 2023-06-13
implemented task

holsoni:
Implemented login() and findByEmail() methods

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = new UserService().findByEmail(email);
+        if (user != null) {
+            if (user.getPassword() == password) {
```
KseniiaMakarova, line 18: should we use == to compare Strings?

KseniiaMakarova (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = new UserService().findByEmail(email);
+        if (user != null) {
+            if (user.getPassword().equals(password)) {
+                return true;
+            }
```
KseniiaMakarova, line 20:
```suggestion
            return user.getPassword().equals(password);
```

KseniiaMakarova (approved):

PR #569, opened 2023-06-14
done functionality to login user, all tests passed

satori11111:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null) {
+            User user = userService.findByEmail(email);
```
nacenik, line 18: You can simplify your solution. You don't need to call the method `userService.findByEmail` twice 
```suggestion
        User user = userService.findByEmail(email);
        if (user != null) {
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (email.equals(users[i].getEmail())) {
+                return users[i];
+            }
+        }
```
nacenik, line 22: Better to use here for-each type of the loop

nacenik (changes requested):
Good solution! Let's improve it)

PR #571, opened 2023-06-18
implemented a login and findByEmail methods

BlueVioletTeti:
implement a login method in AuthenticationService and findByEmail method in UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -10,7 +10,13 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    UserService userService = new UserService();
     public boolean login(String email, String password) {
-        return false;
+        if (userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password)) {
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,14 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        if (userService.findByEmail(email) != null
```
KseniiaMakarova, line 16: let's use local variable for userService.findByEmail(email)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,14 @@ public class AuthenticationService {
...
+                && userService.findByEmail(email).getPassword().equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
KseniiaMakarova, line 21:
```suggestion
        } 
        return false;
```

KseniiaMakarova (changes requested):

PR #573, opened 2023-06-20
implemented task

OlegTuryk:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+
```
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
...
+        for (User user : users) {
+            if (user.getEmail().equals(email)) {
+                return new User(user.getEmail(), user.getPassword());
```
KseniiaMakarova, line 21:
```suggestion
                return user;
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,17 @@ public class AuthenticationService {
...
+        if (user == null) {
+            System.out.println("Error, no such email address");
+            return false;
+        }
+        if (user.getPassword().equals(password)) {
+            return true;
+        } else {
+            System.out.println("Error, wrong password");
+            return false;
+        }
```
KseniiaMakarova, line 27:
```suggestion
        } 
        System.out.println("Error, wrong password");
        return false;
```

KseniiaMakarova (changes requested):

KseniiaMakarova (approved):

PR #576, opened 2023-06-20
implemented task

OlehReex:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@ public class UserService {
...
+            if (email.equals(user.getEmail())) {
+                return user;
+            }
+        }
         return null;
     }
+
+    public static User[] getUsers() {
+        return users;
+    }
```
Andriy-dz, line 28: I don't think you need the getUsers method here

Andriy-dz (approved):

PR #577, opened 2023-06-20
'login' and 'findByEmail' methods are implemented

Alexsandr-Yablunovskyi:
'login' and 'findByEmail' methods are successfully implemented.
The required output in consol is received

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -11,6 +11,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if(UserService.findByEmail(email) != null) {
+            return UserService.findByEmail(email).getPassword().equals(password);
+        }
         return false;
     }
 }
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -14,7 +14,10 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
+        for (User user : users) {
+            if(user.getEmail().equals(email)) return user;
+        }
         return null;
     }
 }


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
```
```suggestion
    private static final User[] users = new User[] {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,16 @@ public class AuthenticationService {
...
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != null) {
+            return UserService.findByEmail(email).getPassword().equals(password);
+        }
```
Andriy-dz, line 17: in this case, you call the findByEmail method twice with the same parameters. I think it is better to create a local change, because I can inject it into performance

Andriy-dz (changes requested):
Good Job! Let's improve your solution)

src/main/java/mate/academy/service/UserService.java
```diff
@@ -10,11 +10,17 @@ public class UserService {
 
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
Andriy-dz, line 18: let's not change the signature of the method (remove the static keyword) and simply create a variable for the UserService class in AuthenticationService

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User verifiableUser = UserService.findByEmail(email);
+        if (verifiableUser != null) {
+            return verifiableUser.getPassword().equals(password);
+        }
         return false;
```
Andriy-dz, line 21: this code can be written in one line

Andriy-dz (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
Alexsandr-Yablunovskyi: How it can be written in one line. You mean like this:
if (verifiableUser != null) return verifiableUser.getPassword().equals(password);

But what about Java Guide Style requirement then?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,21 @@
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService testUser = new UserService();
```
Andriy-dz, line 17: it's not a testUser. it's userService)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,21 @@
...
+        User verifiableUser = testUser.findByEmail(email);
```
Andriy-dz, line 18: User not yet verified. I recommend calling it simply 'User'

Andriy-dz (approved):

PR #579, opened 2023-06-21
Tests passed

Zel1oy:


KseniiaMakarova (changes requested):
- fix code style
- use local variables where possible
- use proper variable names (users -> userService) and so on

PR #585, opened 2023-06-23
Updated

Zel1oy:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) == null) {
```
KseniiaMakarova, line 15: let's use local variable for userService.findByEmail(email)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
...
+            return false;
+        }
+        if (email.equals(userService.findByEmail(email).getEmail())
+                && password.equals(userService.findByEmail(email).getPassword())) {
+            return true;
+        }
         return false;
```
```suggestion
        return email.equals(user.getEmail());
```

KseniiaMakarova (changes requested):

PR #591, opened 2023-07-03
implemented_all_methods

Oleksandr-Tymoshenko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+        if (user != null) {
+            if (user.getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
```
Andriy-dz, line 22:
```suggestion
       return user != null && user.getPassword().equals(password);
```

Andriy-dz (approved):
Good job! Approve

PR #594, opened 2023-07-10

mmosii:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (User user : UserService.getUsers()) {
```
Andriy-dz, line 16: let's try to limit ourselves to using only the findByEmail method. and don't forget to remove extra empty lines

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
```
Andriy-dz, line 11: I don't think you need the getUsers method here

Andriy-dz (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
Andriy-dz, line 17: let's not change the signature of the method (remove the static keyword) and simply create a variable for the UserService class in AuthenticationService

Andriy-dz (changes requested):

Andriy-dz (approved):

PR #596, opened 2023-07-10
Implemented logging method and findByEmail method

ystankevych:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,27 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+
```
Andriy-dz, line 6: an extra empty line

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,27 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
```
```suggestion
    private static final User[] users = new User[] {
```

PR #597, opened 2023-07-14
Add implementations for the UserSevice, AuthenticationService classes

MinoDentori:
Add a solution for the Java Lesson 0 task. Add implementations for the UserSevice, AuthenticationService classes.

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User attemptingUser = userService.findByEmail(email);
+        if (attemptingUser == null) {
+            return false;
+        }
+        return attemptingUser.getPassword().equals(password);
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
kshuryhin, line 16: I would recommend to make `userService` a class member rather than local variable, as it improves flexibility and simplifies testing

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User attemptingUser = userService.findByEmail(email);
```
kshuryhin, line 17: Generally it is not wrong, however from my perspective, it is more conventional to name this variable like `user` or `foundUser`.

kshuryhin (changes requested):
Good job, some minor comments related to code organisation

src/main/java/mate/academy/service/AuthenticationService.java
MinoDentori: Fixed. Is it a good practice in general, to inject other classes in this way, or just in this case?

MinoDentori:
Fix comments.

src/main/java/mate/academy/service/AuthenticationService.java
d1sam: Yes, we will learn about this in the future lectures.

d1sam (approved):
Good job!

PR #599, opened 2023-07-14
Jv oop solution 0

dmytrocherepov:


Andriy-dz (changes requested):
Good job! Let`s improve you solution)

PR #600, opened 2023-07-15
Hm oop

don-bigdad:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (User user : UserService.getUsers()) {
+            if (email.equals(user.getEmail()) && password.equals(user.getPassword())) {
+                return true;
+            }
+        }
```
Andriy-dz, line 20: let's try to limit ourselves to using only the findByEmail method

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,31 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
```
Andriy-dz, line 11: I don't think you need the getUsers and findByPassword methods here

Andriy-dz (changes requested):
Good Job! Let's improve your solution)

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,13 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+
+    public static User findByEmail(String email) {
```
Andriy-dz, line 18: let's not change the signature of the method (remove the static keyword) and simply create a variable for the UserService class in AuthenticationService and delete the regular empty line

Andriy-dz (changes requested):

Andriy-dz (approved):

PR #603, opened 2023-07-19
implemented findByEmail method and login method

pastcwdt:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
```
```suggestion
    private static final User[] users = new User[] {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
...
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        if (userService.findByEmail(email) == null) {
+            return false;
+        } else {
+            return userService.findByEmail(email).getPassword().equals(password);
+
+        }
```
Andriy-dz, line 22: in this case, you call the findByEmail method twice with the same parameters. I think it is better to create a local change, because I can inject it into performance

pastcwdt:
Done!

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,27 @@
...
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        } else {
+            return user.getPassword().equals(password);
+
+        }
```
Andriy-dz, line 25: this code can be written in one line

Andriy-dz (approved):

PR #605, opened 2023-07-19
add methods login and findByEmail 2

TetianaDenchyk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,7 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService member = new UserService();
+        return member.findByEmail(email) != null && member.findByPassword(password) != null;
```
Andriy-dz, line 17: is the logic here correct?) Can there be such a situation that the copier did not enter his password, but that password exists in the database with another user. In this case, your program will skip the user and it will not be very true. I suggest limiting yourself to one getByEmail method and getting the password from the User and comparing it with what was entered at the entrance)

Andriy-dz (changes requested):
Good job! Let's impove your solution)

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
...
      */
     public User findByEmail(String email) {
+        for (User user : users) {
+            if (user.getEmail().equals(email)) {
+                return user;
+            }
+        }
         return null;
     }
+
```
Andriy-dz, line 25: extra empty line

Andriy-dz (approved):

PR #610, opened 2023-07-27
implemented methods of entrance

sonyasapsan:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService temp = new UserService();
+        if (temp.findByEmail(email) != null) {
+            if (temp.findByEmail(email).getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
```
Andriy-dz, line 20: we can write this code in one line. Also, let's make a local variable for the result of the findByEmail method so as not to call it twice - this can affect the performance of the program

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    User findByEmail(String email) {
```
Andriy-dz, line 17:
```suggestion
    public User findByEmail(String email) {
```

Andriy-dz (changes requested):
Good job! Let's improve your solution)

src/main/java/mate/academy/service/AuthenticationService.java
sonyasapsan: done 🤝🤝

Andriy-dz (approved):

PR #625, opened 2023-08-03

Vt0rtikV:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
...
      * Return false in any other cases.
      */
+    private UserService service = new UserService();
+
     public boolean login(String email, String password) {
+        User user = service.findByEmail(email);
+        if (user != null) {
+            return password.equals(user.getPassword());
+        }
         return false;
```
Andriy-dz, line 22: return user != null && password.equals(user.getPassword());

Andriy-dz (approved):

PR #627, opened 2023-08-05

yevhenkukharchuk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService service = new UserService();
+        User user = service.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
```suggestion
        return user != null && user.getPassword().equals(password);
```
You can simplify it ;)

liliia-ponomarenko (approved):

PR #631, opened 2023-08-29
updated login and findByEmail methods

fedorovychh:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,14 @@ public class AuthenticationService {
...
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
kshuryhin, line 23:
```suggestion
     return user != null && user.getPassword().equals(password);
```

kshuryhin (approved):
Great job! Left suggestion which will just improve code readability here

src/main/java/mate/academy/service/AuthenticationService.java
proxychang: thanks, looks really more clear

PR #635, opened 2023-09-05
implement methods "findByEmail" and "login"

dima-semeniuk:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,17 @@
...
             new User("bob@i.ua", "1234"),
             new User("alice@i.ua", "1234")
     };
 
-    /**
-     * Find user by email. All users are stored in <code>private static final User[] users</code>
-     * @param email - the input parameter
-     * @return - user if his email is equal to passed email.
-     * Return <code>null</code> if there is no suitable user
-     */
```
kshuryhin, line 16: Java documentation is crucial when reading somebody's else code, so I would strongly recommend not to remove it

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,17 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static User[] users = new User[] {
```
kshuryhin, line 6: Just wondering what is the reason for removing `final` modifier here?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,12 @@
...
 public class AuthenticationService {
-    /**
-     * Imagine that some user wants to login to your site.
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
```
kshuryhin, line 12: Same about `javadoc`, please do not remove it

src/main/java/mate/academy/Main.java
```diff
@@ -3,7 +3,7 @@
 import mate.academy.service.AuthenticationService;
 
 public class Main {
-    private static final AuthenticationService authenticationService = new AuthenticationService();
+    private static AuthenticationService authenticationService = new AuthenticationService();
```
kshuryhin, line 6: According to the task, you should not modify the `Main` class

kshuryhin (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
+            if (user.getEmail() != null && user.getEmail().equals(email)) {
```
kshuryhin, line 19: I would say here check if user email is null is kinda not needed, however it could be a case for some other inputs.

kshuryhin (approved):

PR #636, opened 2023-09-10
added 'login' and 'findByEmail' methods

HovorukhaBohdan:
…d 'UserService' classes

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -11,6 +11,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService checkUser = new UserService();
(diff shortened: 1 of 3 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService checkUser = new UserService();
```
kshuryhin, line 16: I would recommend to make `UserService` a class member, as it will improve flexibility and simplify testing. Also, consider more appropriate variable name for the service, like `userService`.

kshuryhin (changes requested):
Great job! Just some organisational stuff to consider

kshuryhin (approved):

PR #637, opened 2023-09-14
implemented 'findByEmail' and 'login' methods

YevheniiVlasenko:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
+        return users;
+    }
```
kshuryhin, line 13: Not sure if we really need here such sort of getter. Generally, the idea of the task was to build `AuthenticationService` on top of `UserService`, hence `users` should not be accessible directly from `login` method.

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
...
+
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user : UserService.getUsers()) {
```
kshuryhin, line 22: Here we can access our `users` field directly, as it is in the same class.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (User user : UserService.getUsers()) {
+            if (user.getEmail().equals(email) && user.getPassword().equals(password)) {
+                return true;
+            }
+        }
```
kshuryhin, line 20: Please, refactor this method in the way it will use `UserService` for fetching user by email.

kshuryhin (changes requested):
Nice try, but consider some refactoring just to make sure our `UserService` is doing it's job

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,11 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        if (userService.findByEmail(email)  != null) {
+            return userService.findByEmail(email).getPassword().equals(password);
+        }
         return false;
```
d1sam, line 21: Do you really need to call userService.findByEmail(email) twice?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    UserService userService = new UserService();
```
d1sam, line 6: Could you add access modifier?

d1sam (changes requested):
Some minor changes should be done. Check my comments below :)

src/main/java/mate/academy/service/AuthenticationService.java
YevheniiVlasenko: I don't get it. To which object? Both user credentials and "users" array are private.

d1sam (approved):
Well done 👍

PR #639, opened 2023-09-17
Implemented methods in Java oop lesson0

Zakhar27:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -2,19 +2,28 @@
 
 import mate.academy.model.User;
 
-public class UserService {
+public class UserService extends User {
```
kshuryhin, line 5: That's pretty cool you've managed to achieve some sort of inheritance here. However, I would say this is not the case here. We use inheritance when we have 'is-a' relationship. Here, it would be incorrect to say that our `UserService` is `User`.
You can read more [here](https://en.wikipedia.org/wiki/Inheritance_(object-oriented_programming))

src/main/java/mate/academy/service/UserService.java
```diff
@@ -2,19 +2,28 @@
...
     private static final User[] users = new User[] {
             new User("bob@i.ua", "1234"),
             new User("alice@i.ua", "1234")
     };
 
+    public User[] getUsers() {
+        return users;
+    }
```
kshuryhin, line 13: I would say such sort of getter is not needed in our situation.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,6 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+public class AuthenticationService extends UserService {
```
kshuryhin, line 3: Same here: inheritance is not the case, because for sure we cannot say that `AuthenticationService` is `UserService`.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (int i = 0; i < getUsers().length; i++) {
+            if (getUsers()[i].getEmail().equals(email)
+                    && getUsers()[i].getPassword().equals(password)) {
+                return true;
+            }
+        }
```
kshuryhin, line 19: I would suggest make use of `UserService.findByEmail` when implementing login method. We should not access users here directly or via getter.

kshuryhin (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,14 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        User userEmail = userService.findByEmail(email);
```
d1sam, line 19: Do you really need to create UserService() on each call of login or should we change declaration level for this variable?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,14 @@ public class AuthenticationService {
...
+        if (userEmail != null) {
+            return userEmail.getPassword().equals(password);
+        }
         return false;
```
d1sam, line 23: Can we simplify this code?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
```
d1sam, line 18:
```suggestion
        for (User user : users) {
```

d1sam (changes requested):
Some changes should be done. Please pay attention to level of variables' declaration.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,11 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User userEmail = userService.findByEmail(email);
```
kshuryhin, line 19: I would say `userEmail` is not correct variable name here. It would be more suitable to do just `user`.

kshuryhin (approved):
Goog job!

PR #640, opened 2023-09-19
implemented task

AgroFix:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService loginCheck = new UserService();
```
kshuryhin, line 16: I would recommend to make UserService a class member. It will improve flexibility and simplify testing. Also, please consider better name, like `userService`.

kshuryhin (changes requested):
Great job! Just some code organisational points to consider

d1sam (approved):
Good job!

PR #641, opened 2023-09-19
implemented findByEmail in UserService; implemented login in Authenti…

DEz-1996:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (users[i].getEmail().equals(email)) {
+                return users[i];
+            }
+        }
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
kshuryhin, line 17: Not sure if we really need this method to be `static`

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != null) {
```
kshuryhin, line 14: I would strongly recommend to make `UserService` a class member. It will improve testing and in general is more flexible approach

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@ public class AuthenticationService {
...
+            if (UserService.findByEmail(email).getPassword().equals(password)) {
```
kshuryhin, line 15: Also consider to fetch user by email once at the beginning. Following this, you won't need to call `findByEmail` method each time you need to conduct some checks.

kshuryhin (changes requested):
Nice try, but consider some refactoring in order to achieve better code quality

DEz-1996:
I hope I correct understand your recommendation to make UserService a class member

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,19 @@
...
-     * @param password - user's password
+     * @param inputtedPassword - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
-    public boolean login(String email, String password) {
-        return false;
+    public boolean login(String email, String inputtedPassword) {
+        User userByEmail = new UserService().findByEmail(email);
+        return userByEmail.getPassword().equals(inputtedPassword);
```
d1sam, line 17: What happens when an email arrives that is not in the users' list? Also think about `new UserService()` on every call `login` method, is there a good solution?

d1sam (changes requested):
Please think about situation when passed email is not in array of users.

d1sam (changes requested):
Please think about situation when passed email is not in array of users.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -2,15 +2,20 @@
...
-     * Imagine that some user wants to login to your site.
+     * Imagine that some user wants to log in to your site.
      * You should check if user credentials (login and password) are valid or not.
      * All users are stored in <code>UserService</code> class.
      * @param email - user's email
-     * @param password - user's password
+     * @param inputtedPassword - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
```
mnyshenko, line 12: there is no need to change these lines

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -2,15 +2,20 @@
...
-    public boolean login(String email, String password) {
+    private final UserService userService = new UserService();
+
+    public boolean login(String email, String inputtedPassword) {
+        if (userService.findByEmail(email) != null && inputtedPassword != null) {
+            return inputtedPassword.equals(userService.findByEmail(email).getPassword());
+        }
```
mnyshenko, line 18: In this case, you call userService.findByEmail(email) twice which is insufficient, because you'll have to iterate over all users twice. It's better to extract this expression to a separate variable

mnyshenko (changes requested):
Critical point to consider

PR #642, opened 2023-09-20
jv-oop-lesson-0

TarasYurkovskyi:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
kshuryhin, line 16: I would recommend to make `UserService` a class member rather than local variable. It will have some benefits related to testing and flexibility.

kshuryhin (changes requested):
Great job! Just one recommendation related to the scope of `UserService`

kshuryhin (approved):
Good job

PR #650, opened 2023-09-22
methods in AuthenticationService and UserService classes work now

Alexander1iubchenko:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
kshuryhin, line 17: Not sure if we really need `static` modifier here.

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
...
+        for (int i = 0; i < users.length; i++) {
+            if (email.equals(users[i].getEmail())) {
+                return users[i];
+            }
+        }
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +12,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != null) {
```
kshuryhin, line 15: I would strongly recommend to make `UserService` a class member. It will improve flexibility and simplify testing. Also consider fetching user by email once at the beginning, hence we will not need to call `findByEmail` method each time we need to conduct some checks.

kshuryhin (changes requested):
Nice try, but consider some refactoring in order to achieve better code quality.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private static final UserService userService = new UserService();
```
d1sam, line 6: Why do you make `UserService` as a `static` field? Do you really need it?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        User found = userService.findByEmail(email);
+        if (found != null) {
+            return found.getPassword().equals(password);
+        }
```
d1sam, line 21: Think about better name for variable with name `found`, please :)
Also you should think about simplifying of `if` condition and its body. I believe you will do the best 👍

d1sam (changes requested):
Good job, but a few minor fixes should be done. Please check my comments.

kshuryhin (changes requested):
Just remind you about existing comments :)

d1sam (approved):
Well done!

PR #651, opened 2023-09-22
added the implementation

flekros:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,17 @@
...
 public class AuthenticationService {
-    /**
-     * Imagine that some user wants to login to your site.
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
```
kshuryhin, line 12: I would recommend not to remove java documentation, as it generally helps when reading somebody's else code

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,17 @@
...
+
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User[] getUsers = userService.getUsers();
+        for (User user : getUsers) {
+            if (user.getEmail() == email && user.getPassword() == password) {
+                return true;
+            }
+        }
```
kshuryhin, line 14: I would recommend to make use of `UserService` when implementing login logic here. Also, consider making it class member, rather than local variable.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,15 @@ public class AuthenticationService {
...
      */
+    private UserService userService = new UserService();
+    private User[] users = userService.getUsers();
+
     public boolean login(String email, String password) {
+        for (User user : users) {
+            if (email.equals(user.getEmail()) && password.equals(user.getPassword())) {
+                return true;
+            }
+        }
```
kshuryhin, line 23: What I meant here, is that we should here use our `userService` for user fetching. We do not need again traverse an array of users.

kshuryhin (changes requested):
Good job, but still some points to consider

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,28 @@
...
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
             new User("bob@i.ua", "1234"),
             new User("alice@i.ua", "1234")
     };
 
+    public User[] getUsers() {
+        return users;
+    }
```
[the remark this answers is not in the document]
kshuryhin: Also, getter here is still not needed, as in `AuthenticationService` we do not need to access array of users

kshuryhin (changes requested):
The only point left is that we need to use our `userService` to fetch user and conduct checks. Also, do not forget to remove getter from `UserService`

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,12 @@ public class AuthenticationService {
...
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        if (userService.findByEmail(email) == null) {
+            return false;
+        }
+        return userService.findByEmail(email).getPassword().equals(password);
```
kshuryhin, line 19: I would recommend to fetch user at the beginning, just to avoid calling `findByEmail` each time any check is conducted

kshuryhin (approved):
Great job! Just some clean up is needed :)

PR #652, opened 2023-09-22
completed all tasks

BraiXen-rar:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,19 @@ public class UserService {
...
+                break;
+            } else {
+                result = null;
+            }
+        }
+        return result;
+    }
+
+    public User[] getUsers() {
+        return users;
```
Ivan95kos, line 31: so difficult to read
```suggestion
    public User findByEmail(String email) {
        for (User user : users) {
            if (user.getEmail().equals(email)) {
                return user;
            }
        }
        return null;
    }
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,18 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        User[] users = userService.getUsers();
```
Ivan95kos, line 18: use the `UserService` and method `findByEmail`

Ivan95kos (changes requested):

PR #653, opened 2023-09-22
Implemented a login method in AuthenticationService and findByEmail method in UserService.

ponomvrenko:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,13 @@ public class UserService {
...
      */
     public User findByEmail(String email) {
-        return null;
+        User result = null;
+        for (int i = 0; i < users.length; i++) {
+            if (email.equals(users[i].getEmail())) {
+                result = users[i];
+                break;
+            }
+        }
```
kshuryhin, line 24: Feel free to use _for-each_ loop here. Also, you can simply return result when you find it rather than `break` the loop.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
kshuryhin, line 16: I would recommend to make `userService` a class member, rather than local variable. It will improve flexibility and simplify testing.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user != null) {
+            return user.getEmail().equals(email) && user.getPassword().equals(password);
```
kshuryhin, line 19: I would say this check on email equality is not needed. We already know, that user here will be with correct email, because it was found by `userService.findByEmail` method.

kshuryhin (changes requested):
Great job! Consider some code improvements

src/main/java/mate/academy/service/UserService.java
ponomvrenko: Thanks for remark! It will be more convenient and easier than my code.

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,25 @@
...
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        User result = null;
+        for (User user : users) {
+            if (email.equals(user.getEmail())) {
+                result = user;
+            }
+        }
```
kshuryhin, line 24: Cool, looks much better. One more thing to add - we can return result on the line 22, just when we find it. Hence we will not need to process the rest of the array, because we already found our user.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = new UserService().findByEmail(email);
```
kshuryhin, line 17: I would recommend to make `userService` a class member, because it improves flexibility and simplifies testing.

kshuryhin (changes requested):
Very good, just minor points left

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
...
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
+    private static final UserService userService = new UserService();
```
d1sam, line 17: Do we need `static` here?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
...
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
         return false;
```
d1sam, line 24: can we simplify this?

d1sam (changes requested):
Small changes need to bo done. check my comments :)

d1sam (approved):
Good job, but be careful with empty lines!

PR #658, opened 2023-09-22
Realised methods in the shortest way possible

EdenParadise:
^^

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import java.util.Objects;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if ((Objects.equals(email, "bob@i.ua")) || (Objects.equals(email, "alice@i.ua"))) {
+            return Objects.equals(password, "1234");
+        }
+
         return false;
     }
 }
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -1,5 +1,6 @@
 package mate.academy.service;
 
+import java.util.Objects;
 import mate.academy.model.User;
 
 public class UserService {
@@ -15,6 +16,14 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        if (Objects.equals(email, users[0].getEmail())) {
+            return users[0];
+        }
+
+        if (Objects.equals(email, users[1].getEmail())) {
+            return users[1];
+        }
+
         return null;
     }
 }


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +16,14 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        if (Objects.equals(email, users[0].getEmail())) {
+            return users[0];
+        }
+
+        if (Objects.equals(email, users[1].getEmail())) {
+            return users[1];
+        }
```
kshuryhin, line 25: I would say this solution is not flexible. For example, what if we have much more users, then just two?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if ((Objects.equals(email, "bob@i.ua")) || (Objects.equals(email, "alice@i.ua"))) {
+            return Objects.equals(password, "1234");
+        }
+
```
kshuryhin, line 19: I would say here we need to build our `login` implementation on top of the `UserService.findByEmail` method.

kshuryhin (changes requested):
Nice try, but please consider refactor code in order to make use of `UserService` and also consider your `findByEmail` implementation flexible enough.

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +9,23 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
+        return users;
+    }
+
```
kshuryhin, line 15: Actually, we do not need this getter here. We should use only `findByEmail` method, when doing login.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,16 @@ public class AuthenticationService {
...
     public boolean login(String email, String password) {
+        for (int i = 0;i < UserService.getUsers().length;i++) {
+            if (Objects.equals(UserService.getUsers()[i].getEmail(), email)) {
+                if (Objects.equals(UserService.getUsers()[i].getPassword(), password)) {
+                    return true;
+                }
+            }
+        }
         return false;
```
kshuryhin, line 24: I would strongly recommend to use `userService.findByEmail` for fetching user by email. Then, you can simply check password against fetched user.

kshuryhin (changes requested):
Great, but make sure you use your services efficient enough

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +15,13 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+
+    public static User findByEmail(String email) {
```
d1sam, line 19: Don’t use static methods in your solution
Static methods are in general a bad practice. Let’s better create an instance of a class which method you want to call.
```suggestion
    public User findByEmail(String email) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,11 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != null) {
+            return Objects.equals(UserService.findByEmail(email).getPassword(), password);
+        }
```
d1sam, line 19: You can simplify your code by `findByEmail` only once, make a variable with type User and then check password from this variable with provided password. Also avoid static methods.

d1sam (changes requested):
Good job, but some changes should be done. Please see my comments below and be careful with `static` methods and :)

kshuryhin (changes requested):
Still there are comments to resolve, please, pay attention

PR #660, opened 2023-09-23
Hw jv oop lesson 0

AdamMudrak:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,25 @@
...
      */
     public User findByEmail(String email) {
-        return null;
+        User emptyUser = null;
+        for (User user : users) {
+            if (user.getEmail().equals(email)) {
+                emptyUser = user;
+            }
+        }
+        return emptyUser;
```
Ivan95kos, line 25:
```suggestion
        for (User user : users) {
            if (user.getEmail().equals(email)) {
                return user;
            }
        }
        return null;
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
...
      * You should check if user credentials (login and password) are valid or not.
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
```
Ivan95kos, line 18: use the `UserService` and method `findByEmail`

Ivan95kos (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
AdamMudrak: I think I've changed everything needed. Check the result out and give me your feedback, would you kindly :)

Ivan95kos (approved):
Nice

PR #662, opened 2023-09-23
fix hw-5

BraiXen-rar:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        for (User user : users) {
+            if (email.equals(user.getEmail())) {
+                return user; // Возвращаем пользователя, если email совпал
+            }
+        }
+        return null; // Возвращаем null, если не нашли пользователя
```
okuzan, line 23: it's fine to have comments to make navigating the code for yourself easier, but please remove them before committing

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,14 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        User[] users = userService.getUsers();
+        for (User user : users) {
+            if (user.getEmail().equals(email) && user.getPassword().equals(password)) {
```
okuzan, line 20: image
you already have this logic elsewhere :) Try to reuse it

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@ public class UserService {
...
+    }
+
+    public User[] getUsers() {
+        return users;
```
okuzan, line 27: findByEmail should be enough

okuzan (changes requested):

olekskov (approved):

PR #664, opened 2023-09-23
addes news methods login in AuthenticationService, findByEmail in Use…

YuriiTrokhymenko:
…rService, length line > 100 in Main

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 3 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
sokimaaa, line 16: UserService should be instantiated once and invoked multiple times.

With the current implementation, UserService is instantiated each time the method is invoked.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user != null) {
+            return (user.getPassword().equals(password));
+        }
         return false;
```
sokimaaa, line 21: condition can be compressed 

```suggestion
        return user != null && user.getPassword().equals(password)
```

sokimaaa (changes requested):

Rommelua (approved):

PR #666, opened 2023-09-24
-changed method findByEmail used stream;

yaroslav-ometiukh:
-added UserService to AuthenticationService to find user by email. 
-compare passwords;


Ivan95kos (approved):
Great job

PR #667, opened 2023-09-24
added email checking and login

DevIgork:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (email.equals(users[i].getEmail())) {
+                return users[i];
+            }
+        }
```
kshuryhin, line 22: Feel free to use _for-each_ loop here

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = new UserService().findByEmail(email);
```
kshuryhin, line 16: I would recommend to have a `UserService` as a class member. It will improve flexibility and simplify testing.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
...
+        try {
+            if (user.getPassword().equals(password)) {
+                return true;
+            }
+        } catch (NullPointerException e) {
```
kshuryhin, line 21: That's cool you've managed to use exception handling here. However, generally it is a bad practice to use exceptions as control flow as it can lead to confusing and hard-to-maintain code. Instead of catching a `NullPointerException`, it's better to explicitly check if `user` is `null`

kshuryhin (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User userService : users) {
```
kshuryhin, line 18:
```suggestion
        for (User user : users) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User userService = new UserService().findByEmail(email);
```
kshuryhin, line 16: First of all, it should be `user` variable name. Also, please make `UserService` instance a class member, as it brings a lot of benenfits.

kshuryhin (changes requested):
Cool! Just some minor stuff to consider

src/main/java/mate/academy/service/AuthenticationService.java
DevIgork: I didn't quite understand the 2nd part, I hope I did as you asked

src/main/java/mate/academy/service/AuthenticationService.java
DevIgork: I didn't quite understand the 2nd part, I hope I did as you asked

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
...
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        User loggedInUser = userService.findByEmail(email);
+        if (loggedInUser != null && loggedInUser.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
d1sam, line 22: Can we simplify this?

d1sam (changes requested):
Please check my comment below and simplify your expression :)

src/main/java/mate/academy/service/AuthenticationService.java
DevIgork: no problem

d1sam (approved):
Good job!

PR #670, opened 2023-09-24
fix all methods

DPotapov93:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (email.equals(users[i].getEmail())) {
+                return users[i];
+            }
+        }
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if ((email.equals("bob@i.ua") && password.equals("1234"))
+                || (email.equals("alice@i.ua") && password.equals("1234"))) {
+            return true;
+        }
```
kshuryhin, line 17: Please, consider refactoring `login` implementation in the way it will use `findByEmail` for user fetching.

kshuryhin (changes requested):

d1sam (approved):
Good job! But think about simplifying!

PR #672, opened 2023-09-24
Added verification methods

forzen123212:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != null
```
kshuryhin, line 14: I would strongly recommend to use `UserService` as class member of `AuthenticationService`. Also, I would say you should firstly fetch user like
`User user = userService.findByEmail(email);`
And after that do all checks.

kshuryhin (changes requested):
Nice try, some comments related to code organization

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        return userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password);
```
d1sam, line 17: Do you really need to call `userService.findByEmail(email)` twice ? You can extract `user` from this method once and then use it.

d1sam (changes requested):
please think about the number of calls to the findByEmail method )

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null) { return user.getPassword().equals(password); }
         return false;
+
```
d1sam, line 21: You can simplify and prettify these lines of code.

PR #674, opened 2023-09-24
implemented login and findByEmail

denysokhrymovych:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,18 @@ public class AuthenticationService {
...
+            new User("alice@i.ua", "1234")
+    };
+
     public boolean login(String email, String password) {
+        for (User user : users) {
+            if (user.getEmail().equals(email) && user.getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
```
kshuryhin, line 27: I believe the desired solution here is to leverage `UserService`, which is already created and it's `findByEmail` method.

kshuryhin (changes requested):
Nice try, but please try to use `UserService` when implementing authentication logic

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
d1sam, line 15: Is it a good choice to create `new UserService();` on every call of method `login`?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,12 @@ public class AuthenticationService {
...
+        if (userService.findByEmail(email) != null) {
+            return userService.findByEmail(email).getPassword().equals(password);
+        }
         return false;
```
d1sam, line 19: Can you simplify this?

d1sam (changes requested):
Please think about creation of `new UserService` and simplify check `user != null` and equals of passwords.

PR #675, opened 2023-09-25
Hw 2

annton-shakur:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        User testUser = userService.findByEmail(email);
```
kshuryhin, line 18: Probably we can just use `user` variable name here, but that's minor

kshuryhin (approved):
Great job!

src/main/java/mate/academy/service/AuthenticationService.java
annton-shakur: Sure, thank you!

PR #677, opened 2023-09-25
did homework

BohdanTrue:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
d1sam, line 16: creation of `userService` should be on method level? We don't need to create on every call of `login` a `new UserService();`. Think about it please!

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
         return false;
```
d1sam, line 22: Can we simplify this code?

d1sam (changes requested):
Some changes should be done. Please pay attention to class-level and method-level variables.

src/main/java/mate/academy/service/AuthenticationService.java
BohdanTrue: Can I make the method from non-static to static to call it from the class?

dshulha (approved):
Good job!

dshulha (approved):
Good job!

dshulha (approved):

d1sam (approved):

PR #679, opened 2023-09-25
Initial commit

gordgelin:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
...
+        UserService userService = new UserService();
+
+        User user = userService.findByEmail(email);
+
+        if (user != null) {
+            if (user.getPassword().equals(password)) {
+                return true;
+            }
+        }
+
```
okuzan, line 25:
```suggestion
           return user != null && user.getPassword().equals(password);           
```

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```

Rommelua (approved):

PR #680, opened 2023-09-25
New methods

petro2061:
in this task i implemented logic of methods according to the condition

### DIFF
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -10,10 +10,12 @@ public class User {
     }
 
     public String getEmail() {
+
         return email;
     }
 
     public String getPassword() {
+
         return password;
     }
 }
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -16,10 +16,8 @@ public class AuthenticationService extends UserService {
 
         User inputUser = new User(email, password);
 
-        if(findByEmail(email)!= null && inputUser.getEmail().equals(email) && inputUser.getPassword().equals(password)){
-
-            return true;
-        }
-        return false;
+        return findByEmail(email) != null
+                && inputUser.getEmail().equals(email)
+                && inputUser.getPassword().equals(password);
     }
 }
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,6 +1,8 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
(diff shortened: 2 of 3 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,26 @@
...
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user :
+                users) {
```
okuzan, line 20:
```suggestion
        for (User user : users) {
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,26 @@
...
+            if (email.equals(user.getEmail())) {
+                return user;
+            }
+        }
         return null;
     }
 }
+
```
okuzan, line 28:
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+
+        User inputUser = new User(email, password);
+
+        return findByEmail(email) != null
+                && inputUser.getEmail().equals(email)
+                && inputUser.getPassword().equals(password);
```
```suggestion
        return inputUser != null
                && inputUser.getPassword().equals(password);
```

okuzan (changes requested):

src/main/java/mate/academy/model/User.java
```diff
@@ -10,10 +10,12 @@ public User(String email, String password) {
     }
 
     public String getEmail() {
+
         return email;
     }
 
     public String getPassword() {
+
         return password;
```
Rommelua, line 19: revert these changes

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,8 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
```
Rommelua, line 5: UserService should be class field

```suggestion
public class AuthenticationService {
```

Rommelua (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
petro2061: I do not understen how i can fix this. What is meen UserService should be class field

Ivan95kos (changes requested):

okuzan (approved):
💪

PR #683, opened 2023-09-26
done

zkolesnyk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,15 @@ public class AuthenticationService {
...
+
+    public AuthenticationService() {
+        userService = new UserService();
+    }
+
     public boolean login(String email, String password) {
-        return false;
+        return Arrays.stream(UserService.getUsers())
+                .anyMatch(user -> user.getEmail().equals(email)
+                        && user.getPassword().equals(password));
```
okuzan, line 24: you can reuse existing logic
}

okuzan (approved):
Good job

PR #685, opened 2023-09-26
My branch

leonidbondarenko69:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
```suggestion
```

Ivan95kos (approved):
Great job, with minor comments

PR #688, opened 2023-09-28
implemented methods:

Watxesnavo:
login in AuthenticationService
findByEmail in UserService


okuzan (approved):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User u : users) {
```
```suggestion
        for (User user : users) {
```

PR #689, opened 2023-09-28
Realized methods findByEmail() in UserService class and login() in Au…

Vlad1507:
…thenticationService class.

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        return userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password);
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        return userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password);
```
Rommelua, line 16: do not call findByEmail method 2 times

Rommelua (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
sokimaaa, line 16: UserService instantiates every time the method is invoked. 

Think about making it class level field.

sokimaaa (changes requested):

sokimaaa (approved):

PR #694, opened 2023-09-30
jv-oop-lesson-0 solution

OlexVoit:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = new User(email, password);
```
d1sam, line 17: You don't need to create User inside `login()` method, because you should get it from `userService.findByEmail()` by provided `email` and check with provided `password`.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
+        UserService userService = new UserService();
```
d1sam, line 18: Think about which variables should be local in the method and which should be class-level.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
+        User compareUser = userService.findByEmail(user.getEmail());
+        if (compareUser != null) {
+            return compareUser.getPassword().equals(user.getPassword());
```
d1sam, line 21: You can simplify these lines of code. Just put your `if` condition `compareUser != null` inside `return` statement.

d1sam (changes requested):
Good job, but there are some minor changes need to be done :)

OlexVoit (commented):
made changes

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,22 @@
...
+    private final UserService userService = new UserService();
+
     /**
-     * Imagine that some user wants to login to your site.
+     * Imagine that some user wants to log to your site.
      * You should check if user credentials (login and password) are valid or not.
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
```
mnyshenko, line 13: there is no need to change these lines

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
...
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
             new User("bob@i.ua", "1234"),
             new User("alice@i.ua", "1234")
     };
 
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
```
mnyshenko, line 13: These changes are also redundant

mnyshenko (approved):
Left a few minor comments, please pay attention to what you change and push

src/main/java/mate/academy/service/AuthenticationService.java
OlexVoit: maven wouldn't let me through and displayed an error. I had to change 'login' to 'log'

PR #695, opened 2023-09-30
finished task

DZarudnyi:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService service = new UserService();
```
```suggestion
```

Ivan95kos (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
DZarudnyi: Can you explain why it should be a class field?

Ivan95kos (approved):
Nice

PR #699, opened 2023-10-01
Task2

slava-moiseev-kramatorsk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,5 +1,9 @@
 package mate.academy.service;
 
+import static mate.academy.service.UserService.users;
+
+import mate.academy.model.User;
+
 public class AuthenticationService {
```
Ivan95kos, line 7: You should use the `UserService` and method `findByEmail` here

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,7 +3,7 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    public static final User[] users = new User[] {
```
Ivan95kos, line 6:
```suggestion
    private static final User[] users = new User[] {
```

Ivan95kos (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
slava-moiseev-kramatorsk: "Your task is to implement a login method in AuthenticationService and findByEmail method in UserService. More details are described in comments over the methods. "  -  this condition in file "README.md".
 I thought I should implement the method body in class, were method create.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService users = new UserService();
+        boolean result = false;
+        for (User user: users.getUsers()) {
```
kozhukhovsky, line 18: use the UserService and method findByEmail here

kozhukhovsky (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService users = new UserService();
```
okuzan, line 14:
```suggestion
        UserService userService = new UserService();
```
you have User model and UserService service, naming shouldn't be confusing

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
...
+        boolean result = false;
+        if (users.findByEmail(email) == null) {
+            return false;
+        }
+        if (password.equals(users.findByEmail(email).getPassword())) {
```
okuzan, line 19: in this impl you search for the user twice. Instead, search once and save the result to a User user local variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
...
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        boolean result = false;
+        if (userService.findByEmail(email) != null
+                && (password.equals(userService.findByEmail(email).getPassword()))) {
+            return true;
+        } else {
+            result = false;
+        }
```
okuzan, line 21: two points not fixed - you should call findByEmail only once, and you don't need to create a local boolean variable, you can return directly

okuzan (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
```
```suggestion
        for (User user : users) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        return user != null && (user.getPassword().equals(password));
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

okuzan (approved):
Well done!

PR #700, opened 2023-10-01
solution added

SSosedova:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user:users) {
```
```suggestion
        for (User user : users) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword() == password) {
```
kozhukhovsky, line 18:
```suggestion
        if (user != null && user.getPassword().equals(password)) {
```

Strings are compared using equals

kozhukhovsky (approved):

PR #709, opened 2023-10-03
added solution

AnnaSaienko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User verifiedUser = new UserService().findByEmail(email);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        if (verifiedUser != null) {
+            return verifiedUser.getPassword().equals(password);
+        } else {
+            return false;
+        }
```
Rommelua, line 21:
```suggestion
            return verifiedUser != null && verifiedUser.getPassword().equals(password);
```

Rommelua (changes requested):

Rommelua (approved):

PR #712, opened 2023-10-04
task done

Gagadu:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Ivan95kos, line 16: `UserService` should be a class field
```suggestion
```

Ivan95kos (changes requested):
Good job overall!

PR #713, opened 2023-10-04
implementation method login and method findByEmail

Volanskyi:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +14,9 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        return (user != null && user.getPassword().equals(password)) ? true : false;
```
okuzan, line 20: Consider simplifying the code. You can directly return the result of the condition instead of using if/else or ternary operation to return true/false. For example:

// Instead of this
if (condition) {
return true;
} else {
return false;
}

// You can do this
return condition;
```suggestion
        return user != null && user.getPassword().equals(password);
```

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
Volanskyi: fixed

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,10 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+
+    private static UserService userService = new UserService();
```
Ivan95kos, line 7:
```suggestion
    private final UserService userService = new UserService();
```

Ivan95kos (approved):
Great job, with minor comment

PR #717, opened 2023-10-05
hw - add methods

NazarOliinyk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,16 @@
...
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
olekskov, line 7: services we rely on should be on class level in order not to create new object each time method is called

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,16 @@
...
+        User userLogin = userService.findByEmail(email);
+
+        if (userLogin != null && password.equals(userLogin.getPassword())) {
+            return true;
+        } else {
+            return false;
+        }
```
olekskov, line 14: simplify:
```suggestion
        return userLogin != null && password.equals(userLogin.getPassword());
```

olekskov (changes requested):

Ivan95kos (approved):
Nice

PR #718, opened 2023-10-06
jv-oop-lesson-0-Korniuk-solution

EugeneKorniuk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService passService = new UserService();
+        return passService.findByEmail(email) != null && passService.findByEmail(email)
+                .getPassword().equals(password);
```
karma-o, line 16: We can save the user that we have found to a variable to avoid calling `findByEmail()` twice

karma-o (changes requested):

PR #720, opened 2023-10-06
lesson_0_added_solution

OlegShvets1:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,15 @@
...
-     * Return false in any other cases.
-     */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
okuzan, line 13: Consider simplifying the code. You can directly return the result of the condition instead of using if/else or ternary operation to return true/false. For example:

// Instead of this
if (condition) {
return true;
} else {
return false;
}

// You can do this
return condition;

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
OlegShvets1: +

Ivan95kos (approved):
Nice

PR #722, opened 2023-10-07
Hw solution jv oop lesson 0

karalexander:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,9 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    
     public boolean login(String email, String password) {
-        return false;
+        User user = new UserService().findByEmail(email);
```
sokimaaa, line 17: UserService class instantiates each time the method is invoked. 

Should be declared as the class-level field to be instantiated once and reused multiple times.

sokimaaa (changes requested):

Rommelua (approved):

PR #723, opened 2023-10-10
Implementation of the password verification method

druhov:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,16 @@
...
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
+
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Ivan95kos, line 8: UserService must be a class field
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,16 @@
...
+        User user = userService.findByEmail(email);
+
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
Ivan95kos, line 14: use ternary operator

Ivan95kos (changes requested):
Good job overall!

PR #724, opened 2023-10-10
solution for jv-oop-lesson-0 task

oleksandr-saulko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,18 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Rommelua, line 17: UserService should be class field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,18 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+
+        if (user == null) {
+            return false;
+        }
+
+        if (user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
Rommelua, line 27:
```suggestion
        return user != null && user.getPassword().equals(password);
```

Rommelua (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
oleksandr-saulko: done

src/main/java/mate/academy/service/AuthenticationService.java
oleksandr-saulko: done, thx for advice ;)

Rommelua (approved):

PR #729, opened 2023-10-13
create function findByEmail in UserService and function login in Auth…

AntonBabychP1T:
…enticationService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -7,17 +7,15 @@ public class AuthenticationService {
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
      * All users are stored in <code>UserService</code> class.
-     *
      * @param email - user's email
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Rommelua, line 17: make UserService  class field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
```suggestion
        return user != null && user.getPassword().equals(password));
```

Rommelua (changes requested):

PR #734, opened 2023-10-16
UserService and AuthenticationService services were realised.

MelnychukMykhailo:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        return userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password);
```
okuzan, line 17: call findByEmail once, save the result of the call to a local variable

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
Rommelua: not fixed

Rommelua (changes requested):

okuzan (approved):
Good job!

PR #737, opened 2023-10-16
All tasks have passed.

sshestaka:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
...
      * Return false in any other cases.
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        User userByEmail = userService.findByEmail(email);
+        if (userByEmail != null && userByEmail.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
okuzan, line 22: Consider simplifying the code. You can directly return the result of the condition instead of using if/else or ternary operation to return true/false. For example:

// Instead of this
if (condition) {
return true;
} else {
return false;
}

// You can do this
return condition;

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
sshestaka: Yes! You are totally right))) I was exhausted)))

olekskov (approved):

PR #741, opened 2023-10-18
added implementation

andriihomenko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (User user: UserService.users) {
```
Rommelua, line 16: do not use a loop in login()

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,7 +3,7 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    protected static final User[] users = new User[] {
```
Rommelua, line 6:
```suggestion
    private static final User[] users = new User[] {
```

Rommelua (changes requested):

Sanderfuck (approved):

PR #743, opened 2023-10-19
final_result

IhorTrokhymchuk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null) {
+            if (userService.findByEmail(email).getPassword().equals(password)) {
```
kozhukhovsky, line 16: It’s better to put it in a variable rather than search twice.

kozhukhovsky (approved):

PR #749, opened 2023-10-23
Added solution to method UserService and AuthenticationService

Sanmartial:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +14,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService service = new UserService();
```
Rommelua, line 17: make UserService class field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +14,12 @@ public class AuthenticationService {
...
+        Optional<User> user = Optional.ofNullable(service.findByEmail(email));
```
Rommelua, line 18: No reason for using Optional here - it makes code more difficult
Optional is useful for method return type (like a warning - we possibly have null inside)

Rommelua (changes requested):

Sanmartial:
Replaced ref to UserService as a class field;
Removed 'Optional'.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
...
      */
+    private final UserService service = new UserService();
+
     public boolean login(String email, String password) {
+
+        User user = service.findByEmail(email);
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
         return false;
```
Sanderfuck, line 24: Simplify it in one line using one return, at first check user != null && condition for email

Sanderfuck (changes requested):

Sanderfuck (approved):

PR #754, opened 2023-10-25
solution complete

Andrewmarkov:
done i hope :) 
Just didn't realise why I couldn't call findByEmail method from AuthenicationService class until I didn't make findByEmail class static (IntelliJ Idea quick fix)

`    public boolean login(String email, String password) {
        if (user != null) {
            return ((User) user).getPassword().equals(password);
        }
        return false;
`

If the method below is not static I can't call it 
`    public static User findByEmail(String email) {
        for (User user : users) {
                return user;
            }
        }
        return null;

Thanks for explanation!
`

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+        if (user != null) {
+            return ((User) user).getPassword().equals(password);
+        }
         return false;
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
Rommelua, line 17:
```suggestion
    public User findByEmail(String email) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
```
Rommelua, line 16: make UserService class field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
...
+        if (user != null) {
+            return ((User) user).getPassword().equals(password);
+        }
         return false;
```
Rommelua, line 20:
```suggestion
            return user != null && user.getPassword().equals(password);
```

Rommelua (changes requested):

sokimaaa (approved):

PR #757, opened 2023-10-27
Implemented a login method in AuthenticationService and findByEmail m…

Khrestina:
…ethod in UserService.

### DIFF
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -14,6 +14,7 @@ public class User {
     }
 
     public String getPassword() {
+
         return password;
     }
 }
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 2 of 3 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -14,6 +14,7 @@ public String getEmail() {
     }
 
     public String getPassword() {
+
```
kerrrusha, line 17:
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,14 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
```
kerrrusha, line 15:
```suggestion
    private final UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,14 @@ public class AuthenticationService {
...
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user != null) {
+            return user.getPassword().equals(password);
+        } else {
+            return false;
+        }
```
kerrrusha, line 23:
```suggestion
        if (user != null) {
            return user.getPassword().equals(password);
         }
        return false;
```
or even
```suggestion
        return user == null ? false : user.getPassword().equals(password);
```

kerrrusha (approved):
see comments

PR #759, opened 2023-10-28
did my homework

Nazar090:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
+        return users;
+    }
```
kerrrusha, line 13: as a Java programmer, your job is to implement only that features that mentioned in the task
image

```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User[] users = UserService.getUsers();
```
kerrrusha, line 16:
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
...
+        for (User user : users) {
+            if (email.equals(user.getEmail())
+                    && password.equals(user.getPassword())) {
+                return true;
+            }
+        }
         return false;
```
kerrrusha, line 23: we've implemented already that logic in the UserService, so why create duplicating code?
use UserService for finding user by email, and do required check with that user

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
kerrrusha, line 14: make this class field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
...
+        while (userService.findByEmail(email) != null) {
+            if (userService.findByEmail(email).getPassword().equals(password)) {
+                return true;
+            } else {
+                return false;
+            }
+        }
         return false;
```
kerrrusha, line 22: why you are using while loop if we have only one user to work with?
save user to the local variable, there's no need to find him via findByEmail() method each time

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null) {
```
kerrrusha, line 15: not fixed previous comment about extracting `userService.findByEmail(email)` to the local variable

src/main/java/mate/academy/service/AuthenticationService.java
kerrrusha: > make this class field

not fixed

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        final User current = userService.findByEmail(email);
+        if (current != null) {
+            return userService.findByEmail(email).getPassword().equals(password);
```
kerrrusha, line 20:
```suggestion
            return current.getPassword().equals(password);
```

kerrrusha (changes requested):

PR #761, opened 2023-10-28
Did all tasks, implemented login() and findByEmail() methods

Trayvor:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,7 +3,7 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    public static final User[] users = new User[] {
```
Ivan95kos, line 6: the field must be the private, revert it
```suggestion
    private static final User[] users = new User[] {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
...
+        for (User user : UserService.users) {
+            if (user.getEmail().equals(email) && user.getPassword().equals(password)) {
+                return true;
+            }
+        }
```
Ivan95kos, line 21: use the method `findByEmail` instead of `loop`

Ivan95kos (changes requested):

okuzan (approved):

PR #762, opened 2023-10-28
Implement a login method in AuthenticationService and findByEmail met…

JaysonJuza:
…hod in UserService.

### DIFF
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -7,13 +7,13 @@ public class UserService {
             new User("bob@i.ua", "1234"),
             new User("alice@i.ua", "1234")
     };
-
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
+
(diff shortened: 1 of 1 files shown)


Sanderfuck (changes requested):
Look what changes you send to PR

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
kerrrusha, line 14: should be class field. no need to create object each time when we are calling this method

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
...
+        return userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password);
```
kerrrusha, line 16: save `userService.findByEmail(email)` to the variable and work with it, to avoid duplicate method calling

kerrrusha (changes requested):

Rommelua (approved):

PR #765, opened 2023-10-29
added login method in AuthenticationService and findByEmail method in…

Andriy:
… UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -16,10 +16,6 @@ public class AuthenticationService {
 
     public boolean login(String email, String password) {
         User user = userService.findByEmail(email);
-        if (user == null) {
-            return false;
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,14 @@ public class AuthenticationService {
...
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        } else {
+            return user.getEmail().equals(email) && user.getPassword().equals(password);
+        }
```
okuzan, line 23:
```suggestion
        return user != null && user.getPassword().equals(password);
        
```

okuzan (changes requested):

kozhukhovsky (approved):

PR #766, opened 2023-10-30

MaksymMazlov:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        return email.equals(user.getEmail()) && password.equals(user.getPassword());
```
sokimaaa, line 23: Merge conditions. 
```suggestion
        if (user == null) {
            return false;
        }
        return email.equals(user.getEmail()) && password.equals(user.getPassword());
```

sokimaaa (changes requested):

Rommelua (approved):

PR #771, opened 2023-10-31
Authentification

rafik2211:
A UserService class is defined, which is responsible for managing user data. A list of users is created as a static field for demonstration purposes, and a getUserByEmail method is implemented to search for a user by their email address. Users with email and password combinations are added to the list as sample data. AuthenticationService:
An AuthenticationService class is defined, which is responsible for user authentication. The login method is implemented to check if the entered email and password are correct for login. The UserService.getUserByEmail method is used to retrieve a user by their email address. If the user is found and the entered password matches the user's password, the method returns true, indicating a successful login. Otherwise, it returns false.

### DIFF
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
+            if (user.getEmail().equals(email)) {
+                return user;
+            }
+        }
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
```
```suggestion
        for (User user : users) {
```

kozhukhovsky (approved):

PR #776, opened 2023-11-02
Lesson 0

MarynaBodina:
Lesson 0


okuzan (approved):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
okuzan, line 23:
```suggestion
        return user != null && user.getPassword().equals(password);
```
just return the statement itself like this, when we are dealing with booleans

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,27 @@
...
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        if (email == null) {
+            return null;
+        }
```
okuzan, line 21:
```suggestion
```
redundant, we have return null at line 27

okuzan (changes requested):

PR #778, opened 2023-11-03
solution for "implement simple authentication"

mykolatsupryk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService service = new UserService();
+        User user = service.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
kerrrusha, line 21: dont complicate if-else statements:
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService service = new UserService();
```
kerrrusha, line 16: should be `private final` class-level field

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService service = new UserService();
```
sokimaaa, line 16: Not fixed prev comment.
Should be class level field. 
```suggestion
        UserService service = new UserService();
```

sokimaaa (changes requested):

sokimaaa (approved):

sokimaaa (approved):

PR #779, opened 2023-11-03
implemented methods

ivanlavrinenko:
login method in AuthentificationService and findByEmail in UserService are implemented

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -3,21 +3,10 @@ package mate.academy.service;
 import mate.academy.model.User;
 
 public class AuthenticationService {
-    /**
-     * Imagine that some user wants to login to your site.
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
+    private final UserService userService = new UserService();
+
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
kerrrusha, line 16: should be class-level `private final` field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        return user.getPassword().equals(password);
```
kerrrusha, line 21: it's ok, but as an extra challenge, you can try to rewrite it in 1 line) if you want, think about how to do it

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
ivanlavrinenko: I have just added new commit on by branch.

ivanlavrinenko: > should be class-level `private final` field

I have just added new commit on by branch, but can't see if it's checked by Java CI.

Rommelua (approved):

PR #780, opened 2023-11-04
task solved

orest40th:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
+        for (User obj: users) {
+            if (obj.getEmail().equals(email)) {
+                return obj;
+            }
+        }
```
sokimaaa, line 22: Dont use static method here
```suggestion
    public User findByEmail(String email) {
        for (User obj: users) {
            if (obj.getEmail().equals(email)) {
                return obj;
            }
        }
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != (null)
+                && UserService.findByEmail(email).getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
sokimaaa, line 18: Use the ternary operator to compress the if condition. 
```suggestion
        if (UserService.findByEmail(email) != (null)
                && UserService.findByEmail(email).getPassword().equals(password)) {
            return true;
        }
        return false;
```

sokimaaa (changes requested):

sokimaaa (changes requested):
see comments

src/main/java/mate/academy/service/UserService.java
orest40th: if i remove the static access modifier here i get a 'Non-static method 'findByEmail(java.lang.String)' cannot be referenced from a static context' problem in my login method. am i missing something?

orest40th:
if i remove the static access modifier here i get a 'Non-static method 'findByEmail(java.lang.String)' cannot be referenced from a static context' problem in my login method. am i missing something?

src/main/java/mate/academy/service/UserService.java
orest40th: @sokimaaa ??

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User obj: users) {
```
Rommelua, line 18:
```suggestion
        for (User user: users) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -2,15 +2,25 @@
...
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
-     */
+    */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        return userService.findByEmail(email) != (null)
+                && userService.findByEmail(email).getPassword().equals(password);
```
Rommelua, line 17: do not call findByEmail 2 times

Rommelua (changes requested):

Sanderfuck (approved):

PR #781, opened 2023-11-05
task done

ArtemMakovskyy:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private static final UserService userService = new UserService();
```
kerrrusha, line 6: services are not constants usually
```suggestion
    private final UserService userService = new UserService();
```

kerrrusha (approved):

PR #783, opened 2023-11-07
solution 1

katrusyavshch:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,17 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Rommelua, line 16: Make UserService a class field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,17 @@ public class AuthenticationService {
...
+
+        if (foundUser == null) {
+            return false;
+        }
+
+        if (foundUser.getPassword() == password) {
+            return true;
+        }
+        
         return false;
```
Rommelua, line 27:
```suggestion
        return foundUser != null && foundUser.getPassword().equals(password);
```

Rommelua (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,11 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        userService = new UserService();
```
Rommelua, line 18: init userService once on line 6 - no reason to do it each time when login method is called

Rommelua (changes requested):

Rommelua (approved):

PR #787, opened 2023-11-09
First commint

VasylPalokha:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
olekskov, line 16: services we rely on better to make class variable in order not to create new object each time method is called

olekskov (approved):
Fine overall, approved, please pay attention to comment in code for future tasks

PR #788, opened 2023-11-09
hm-2-oop

Cyrax977:
make logik to methods AuthenticationService and UserService


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,18 @@ public class AuthenticationService {
...
+    }
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
sokimaaa, line 27: the if statement could be simplified.

sokimaaa (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
Cyrax977: ok thnx. I'll fix code

Cyrax977 (commented):
simplified the if statment

kerrrusha (approved):

PR #790, opened 2023-11-09
complete homework

Kinereiko:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
Rommelua, line 17:
```suggestion
    public User findByEmail(String email) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
...
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User checkUser = UserService.findByEmail(email);
+        if (checkUser == null) {
+            return false;
+        } else {
+            return checkUser.getPassword().equals(password);
+        }
```
```suggestion
            return checkUser != null && checkUser.getPassword().equals(password);
```

Rommelua (changes requested):

Rommelua (changes requested):
see comments

Kinereiko:
Исправил и сделал новый commit updated homework: 7715c8df34d3e1184f287788b4532767669b4c56

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,21 @@
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService serviceUsers = new UserService();
```
kerrrusha, line 17: user service should be `private final` class-level field, no need to create it each time when we call this method
also, consider using simple naming like `userService`

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
Kinereiko: I don't understand how to call method "findByEmail" without modificator "static" and without creating "UserService" instance of a class.

sokimaaa (approved):

PR #791, opened 2023-11-10
AuthenticationService add

artemsa223:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,25 @@
...
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+
+        for (int i = 0; i < users.length; i++) {
+            if (email.equals(users[i].getEmail())) {
+                return users[i];
+            }
+        }
```
kerrrusha, line 24: better use for-each loop

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,25 @@
...
 
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+
```
kerrrusha, line 19: redundant blank line
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,27 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
```
kerrrusha, line 5: this class can't be inherited from UserService. in this case, you should use aggregation

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,27 @@
...
     public boolean login(String email, String password) {
+        User user = findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        if (password.equals(user.getPassword())) {
+            return true;
+        }
+
         return false;
```
kerrrusha, line 25: don't complicate if-else logic. think about how to rewrite it in 1 line

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = use.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        return password.equals(user.getPassword());
```
sokimaaa, line 23: merge conditions, e.g. could be rewrite in one line
```suggestion
        if (user == null) {
            return false;
        }
        return password.equals(user.getPassword());
```

sokimaaa (changes requested):

sokimaaa (approved):

PR #793, opened 2023-11-10
oop first solution

KoderNL:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
...
+        UserService userService = new UserService();
+
+        if (userService.findByEmail(email) != null) {
+            User user = userService.findByEmail(email);
+            if (user.getPassword().equals(password)) {
+                return true;
+            }
+            return false;
+        }
         return false;
```
Sanderfuck, line 25: userService.findByEmail(email) - you can save it to variable and use it many times

Sanderfuck (commented):
image
Fix this

Sanderfuck (changes requested):
Fix problem with checks in github

KoderNL:
> image Fix this

я не розумію в чому тут проблема та як я можу її пофіксити

KoderNL:
I'm already fixed last problem and change some code with your advice, you can check please

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+
+        if (userService.findByEmail(email) != null) {
+            return user.getPassword().equals(password);
+        }
```
Sanderfuck, line 21: You already have user field, why do you duplicate code of user inside if(......)?
You can simplify it to one line in return

Sanderfuck (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
         return false;
```
kerrrusha, line 22:
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
kerrrusha, line 16: services should be `private final` class-level fields
```suggestion
     */
    private final UserService userService = new UserService();
    
    public boolean login(String email, String password) {
```

koval-kn (approved):

PR #795, opened 2023-11-11
implement simple authentication

VitaliiTeben:


sokimaaa (changes requested):
image

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,27 @@
...
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (User confirmUser : users) {
+            if (confirmUser.getEmail().equals(email)
+                    && confirmUser.getPassword().equals(password)) {
+                return true;
+            }
+        }
```
sokimaaa, line 24: Use UserService here, no need to duplicate code.
```suggestion
        for (User confirmUser : users) {
            if (confirmUser.getEmail().equals(email)
                    && confirmUser.getPassword().equals(password)) {
                return true;
            }
        }
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,25 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    public static final User[] users = new User[] {
```
sokimaaa, line 6: Make it private field, cause we dont want to expose our users everywhere
```suggestion
    private static final User[] users = new User[] {
```

sokimaaa (changes requested):

Sanderfuck (approved):

PR #797, opened 2023-11-13
added solution

roman680:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,16 @@
...
+
+    private static UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        return user.getEmail().equals(email) && user.getPassword().equals(password);
```
Sanderfuck, line 14: User what you get from userService.findByEmail(email) already checked email, don`t duplicate! You can check user for null and compare password in one line of return, let think about how

Sanderfuck (changes requested):

Sanderfuck (changes requested):

kerrrusha (approved):

PR #798, opened 2023-11-14
is done

DenysShl:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        return password.equals(user != null ? user.getPassword() : null);
```
kerrrusha, line 18: you've combined two different checks in wrong way (if user provided `null` password and user is `null` - so such user would be logged in). 
consider check for null && check for password equality without ternary expressions, they are redundant here

kerrrusha (changes requested):

PR #803, opened 2023-11-15
Service realizathion

maiiat:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +16,9 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        if (Arrays.stream(users).anyMatch(e -> e.getEmail().equals(email))) {
+            return Arrays.stream(users).filter(e -> e.getEmail().equals(email)).findFirst().get();
+        }
```
Sanderfuck, line 21: Use just loop here

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService user = new UserService();
+        User actualUser = user.findByEmail(email);
+        if (actualUser != null && actualUser.getEmail().equals(email)
+                && actualUser.getPassword().equals(password)) {
```
Sanderfuck, line 19: You checked by email in method findByEmail(email)

Sanderfuck (changes requested):

src/main/java/mate/academy/service/UserService.java
maiiat: Thank you, updated

maiiat:
@Sanderfuck can you please review one more time

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
...
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService user = new UserService();
+        User actualUser = user.findByEmail(email);
+        if (actualUser != null && actualUser.getPassword().equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
Sanderfuck, line 22: You can simplify it when return (condition what inside if) and you don`t need to return true or false

Sanderfuck (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
maiiat: @Sanderfuck updated

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService user = new UserService();
```
sokimaaa, line 16: 1. bad var name for user service
2. declare the service as class level field
```suggestion
        UserService user = new UserService();
```

sokimaaa (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
maiiat: @sokimaaa updated

PR #809, opened 2023-11-21
 add AuthenticationService solution

Svitlankaa:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
-     */
+
+    private static final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
Sanderfuck, line 14: Try to simplify it, return result of condition and you don`t need to write return true and false

Sanderfuck (changes requested):

okuzan (approved):
Well done!

PR #810, opened 2023-11-21
create two methods to login method

RomanSoproniuk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,31 @@
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
okuzan, line 19: make it a private class field, initialize in constructor

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
...
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (email.equals(users[i].getEmail())) {
+                return users[i];
+            }
+        }
         return null;
     }
+
+    public boolean findByPassword(String password, String userPassword) {
```
okuzan, line 20: name doesn't correspond to the actual logic, remove this method, move logic to login method

okuzan (changes requested):
We need to deal with boolean variables slightly different:

You can directly return the result of the condition instead of using if/else or ternary operation to return true/false. For example:

// Instead of this
if (condition) {
return true;
} else {
return false;
}

// You can do this
return condition;

src/main/java/mate/academy/service/UserService.java
RomanSoproniuk: doesn't matter anymore

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
...
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        User newUser = userService.findByEmail(email);
+        if (newUser != null) {
+            return password != null && password.equals(newUser.getPassword());
+        }
         return false;
```
okuzan, line 24:
```suggestion
       return newUser != null && password != null && password.equals(newUser.getPassword());
```

okuzan (approved):

PR #811, opened 2023-11-22
fixed AuthenticationService class

olieniev:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
kerrrusha, line 16: make `userService` as class-level `private final` field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        return (password == user.getPassword());
```
kerrrusha, line 21: 1. we cannot compare strings by `==` operator, google about why we need to use `.equals()` rather
2. we can make it more simple
```suggestion
        return user != null && password.equals(user.getPassword());
```

kerrrusha (changes requested):

kerrrusha (approved):

PR #812, opened 2023-11-22
solution

anastezia12:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,17 @@
...
-    /**
-     * Find user by email. All users are stored in <code>private static final User[] users</code>
-     * @param email - the input parameter
-     * @return - user if his email is equal to passed email.
-     * Return <code>null</code> if there is no suitable user
-     */
     public User findByEmail(String email) {
+        for (User x : users) {
+            if (x.getEmail().equals(email)) {
+                return x;
```
kerrrusha, line 14: very bad naming.. better not to make our code so obfuscated)
```suggestion
        for (User user : users) {
            if (user.getEmail().equals(email)) {
                return user;
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,14 @@
...
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
kerrrusha, line 7: make it `private final` class-level field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,14 @@
...
+        User x = userService.findByEmail(email);
+        if (x == null) {
+            return false;
+        }
+        return x.getPassword().equals(password);
```
kerrrusha, line 12: 1. bad `User` variable naming, never name variables like so except counters or lambdas
2. too complicated if-elfse logic. think about how to rewrite to do it in 1 line
```suggestion
        User user = userService.findByEmail(email);
        if (user == null) {
            return false;
        }
        return user .getPassword().equals(password);
```

kerrrusha (changes requested):

PR #816, opened 2023-11-25
Homework jv-oop-lesson-0 .

UniyatYura:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,17 @@
...
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
olekskov, line 8: Please make class field to avoid creation of new object each time method is called

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,17 @@
...
+        Optional<User> byEmail = userService.findByEmail(email);
+        Optional<User> byPassword = userService.findByPassword(password);
+        if (byEmail.isEmpty() || byPassword.isEmpty()) {
+            return false;
+        }
+        return email.equals(byEmail.get().getEmail())
+                && password.equals(byPassword.get().getPassword());
```
olekskov, line 15: your logic is not correct
you may obtain login from one user and password from another and consider this combination valid
please find user by email only and check if input password equals with this user's password

olekskov (changes requested):

olekskov (approved):
Good job

PR #819, opened 2023-11-29
added implementation of AuthenticationService and findByEmail methods…

Parabellum7:
… in UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,16 +1,27 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
+        return users;
+    }
```
olekskov, line 13: cannot see reasons to create this method

```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,27 @@
...
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User[] users = UserService.getUsers();
+        for (User user : users) {
+            if (user.getEmail().equals(email) && user.getPassword().equals(password)) {
+                return true;
+            }
+        }
```
olekskov, line 23: half of code repeats findByEmail(String email) method
please create class field for UserService and use it's method here

```suggestion
        User users = userService.findByEmail();
        return user.getPassword().equals(password);
```

olekskov (changes requested):

PR #827, opened 2023-12-03
first version

SieryiD:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != null
+                && UserService.findByEmail(email).getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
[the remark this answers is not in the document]
okuzan: Store result of findByEmail in local variable, don't call it twice
method shouldn't be static

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
kerrrusha, line 16: should be `private final` class-level field. no need to create it each time when we call this method

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,9 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
```
kozhukhovsky, line 16:
```suggestion
```

kozhukhovsky (approved):

PR #829, opened 2023-12-04
implemented login

yenaofox:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        return (user != null && user.getPassword().equals(password)) ? true : false;
```
Sanderfuck, line 18:
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Sanderfuck, line 16: Every invoke of method create new object of UserService, you can save it in field of class

Sanderfuck (changes requested):

PR #830, opened 2023-12-04
add findByEmail and login methods

Nikname2303:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,27 @@
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userservice = new UserService();
+        User check = userservice.findByEmail(email);
```
karma-o, line 18: Let's name this var as just user, which it is
```suggestion
        User user = userservice.findByEmail(email);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,27 @@
...
+        if (check == null) {
+            return false;
+        }
+        if (email.equals(check.getEmail()) && password.equals(check.getPassword())) {
+            return true;
+        }
```
karma-o, line 24: we can simplify it to this
```suggestion
        return user != null && user.getPassword().equals(password)
```
since email is basically already checked, and we can invert the null check to fit it inside one single check and return it right away without any `if`s :)

karma-o (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,21 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
-    /**
+    /*****
```
karma-o, line 6: let's make it as it was before :)
```suggestion
    /**
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,21 @@
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userservice = new UserService();
```
karma-o, line 17: sorry for missing it the first time, we should also not create a new `UserService` each time we call login method, but make it a class level variable :)

karma-o (approved):
GJ :+1: 
But please also check an important comment about something that I have missed in the first review

Nikname2303:
ok, now I don't call variable every time, but I am not shure it's right, but now UserService is private final variable

PR #832, opened 2023-12-04
Working with services task is done.

bodya4243:
1. implemented  a "findByEmail" method in the UserService class.
2. implemented a "login" method in the AuthenticationService class.
3. checked code style with command "mvn clean package"

### DIFF
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -16,5 +16,4 @@ public class User {
     public String getPassword() {
         return password;
     }
-
 }
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -15,15 +15,11 @@ public class AuthenticationService {
     private final UserService userService = new UserService();
 
     public boolean login(String email, String password) {
-        try {
-            User approvedUser = userService.findByEmail(email);
-            if (password.equals(approvedUser.getPassword())) {
-                return true;
-            }
-        } catch (NullPointerException e) {
+        User approvedUser = userService.findByEmail(email);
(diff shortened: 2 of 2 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -16,4 +16,5 @@ public String getEmail() {
     public String getPassword() {
         return password;
     }
+
```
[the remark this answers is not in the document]
bodya4243: I tried to delete these empty lines but without it the program isn't building.
Снимок экрана 2024-01-30 164616

src/main/java/mate/academy/model/User.java
```diff
@@ -1,8 +1,8 @@
 package mate.academy.model;
 
 public class User {
-    private String email;
-    private String password;
+    private final String email;
+    private final String password;
```
karma-o, line 5: I can see why you did that, but usually fields in POJO classes are not final unless we want to explicitly make them final and immutable(we will talk about it more during the course), you can leave it as it is, but be careful with that next time :)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,14 @@ public class AuthenticationService {
...
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User approvedUser = userService.findByEmail(email);
+        if (approvedUser != null && password.equals(approvedUser.getPassword())) {
+            return true;
+        } else {
+            return false;
+        }
```
karma-o, line 23: also some minor optimization :wink: 
```suggestion
        return approvedUser != null && password.equals(approvedUser.getPassword());
```

karma-o (approved):
GJ! :+1: 
Also Left some minor comments

PR #833, opened 2023-12-05
added value AuthenticationService UserService

RCherniavskiy:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,6 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+public class AuthenticationService extends UserService {
```
kozhukhovsky, line 3:
```suggestion
public class AuthenticationService {
```

use UserService as variable

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User number: users) {
```
```suggestion
        for (User user : users) {
```

kozhukhovsky (changes requested):

PR #835, opened 2023-12-05
added value AuthenticationService and UserService

RCherniavskiy:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User number: users) {
+            if (email.equals(number.getEmail())) {
+                return number;
```
Sanderfuck, line 20: Number not informative naming, better call it "user"

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (findByEmail(email) != null) {
+            return findByEmail(email).getPassword().equals(password);
+        }
         return false;
```
Sanderfuck, line 17: Simplify it using one return, findByEmail(email) - I see duplication of code, you can save it in variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,6 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+public class AuthenticationService extends UserService {
```
Sanderfuck, line 3: Don`t extends from UserService it`s incorrect

Sanderfuck (changes requested):

PR #837, opened 2023-12-06
added new AuthenticationService and UserService

RCherniavskiy:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService user = new UserService();
+        if (user.findByEmail(email) != null) {
+            return user.findByEmail(email).getPassword().equals(password);
+        }
         return false;
```
```suggestion
            return user != null && user.getPassword().equals(password);
        }
```
don't call findByEmail twice, store the result of the call in a local variable, also, we can combine expressions like this

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
```
```suggestion
        for (User user : users) {
```
make sure to format your files before a review - Ctrl ALt L

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
RCherniavskiy: Condition 'user != null' is always 'true'

PR #839, opened 2023-12-07
add AuthenticationService UserService

RCherniavskiy:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,12 @@
...
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
+
     public boolean login(String email, String password) {
+        UserService user = new UserService();
```
okuzan, line 6: make it a private field instead

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,12 @@
...
+        if (user.findByEmail(email) != null) {
```
okuzan, line 7: store result of this call, don't call this method twice
return one boolean expression

kozhukhovsky (approved):

PR #840, opened 2023-12-08
solved task

AntonKurylo:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
...
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        User findByEmailUser = userService.findByEmail(email);
+        if (findByEmailUser != null
+                && findByEmailUser.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
kerrrusha, line 24: could be simplified into 1 line

sokimaaa (approved):

PR #842, opened 2023-12-08
made authentication

BelitskyiAlexandr:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
sokimaaa, line 20: the code could be simplified
```suggestion
        return /* some merged boolean condition */;
```

sokimaaa (changes requested):

PR #843, opened 2023-12-09
solution to hw-jv-oop-lesson-0

bilberryyy:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,14 @@ public class UserService {
...
+        for (User user : users) {
+
+            if (user.getEmail().equals(email)) {
+
+                return user;
+            }
+        }
+        return null; // Користувача з вказаною електронною адресою не знайдено
     }
+
```
okuzan, line 27:
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,14 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        for (User user : users) {
+
+            if (user.getEmail().equals(email)) {
+
```
okuzan, line 21:
```suggestion
            if (user.getEmail().equals(email)) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+
+        UserService userService = new UserService();
```
okuzan, line 17: make it a private field, initialize in constructor, or at declaration

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+
+        // Check if the user with the given email exists
+        if (user != null) {
+            // Check if the provided password matches the user's password
+            return user.getPassword().trim().equals(password.trim());
+        }
+
         return false;
```
okuzan, line 26: good job, we can also combine these two conditions and return one logical expression instead. Also, trimming is a good idea, but on a service level we usually receive sanitized input and check more complex business rules, so you can omit it in this task

okuzan (changes requested):

PR #844, opened 2023-12-10
solved task

Leanchers:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,20 @@ public class AuthenticationService {
...
      */
     public boolean login(String email, String password) {
+        UserService user = new UserService();
+        if (user.findByEmail(email) != null) {
+            if (user.findByEmail(email).getEmail().equals(email)
+                    && user.findByEmail(email).getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
```
okuzan, line 22: 1. don't create a service object per each method call - make it a class field
2. Don't call UserService object user - it's very confusing since we have model User
3. Save the result of userService.findByEmail to a local variable - your method will be 3 times more efficient :)
4. Try to achieve the same without using boolean literals like true and false :)

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
```
okuzan, line 6: undo this change

okuzan (changes requested):

PR #851, opened 2023-12-14
Implemented simple authentication

davydov-software:
Implemented `login` method in `AuthenticationService` and `findByEmail` method in `UserService`.

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
olekskov, line 16: Currently code will create new UserService each time method is called.
Please move userService to class variable to avoid such situation

olekskov (changes requested):

olekskov (approved):
Good job!

PR #852, opened 2023-12-14
implemented all of the tasks

MolchanovAlexander:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
```
karma-o, line 6: This code section is a Javadoc of the `login()` method and should be right above it, class fields on the other hand should be at the beginning of the class

karma-o (approved):
GJ! Left a style comment

src/main/java/mate/academy/service/AuthenticationService.java
karma-o, line 6: This code section is a Javadoc of the login() method and should be right above it, class fields on the other hand should be at the beginning of the class

karma-o (approved):
GJ! Left a style comment

src/main/java/mate/academy/service/AuthenticationService.java
MolchanovAlexander: image
maven wants me to add  [EmptyLineSeparator]

PR #856, opened 2023-12-16
implemented logi and findByEmail methods

YuliiaNisha:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,13 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+
+    public static User findByEmail(String email) {
```
olekskov, line 18: Please do not use static methods unless necessary
```suggestion
    public User findByEmail(String email) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = UserService.findByEmail(email);
```
olekskov, line 16: instead of static method call make class variable of UserService type and use it to call method

olekskov (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
```suggestion
```

Ivan95kos (changes requested):
Good job overall!

okuzan (approved):
Well done

PR #858, opened 2023-12-17
chenged login and findByEmail methods

SerhiiAN:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
okuzan, line 16: make it a private class field instead

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
```
okuzan, line 20: try to rewrite this method without using `false` literal

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,13 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        User findUser = null;
+        for (User user : users) {
+            if (user.getEmail().equals(email)) {
+                findUser = user;
+                break;
```
okuzan, line 22: more usual approach would be to return user here, and null at the end of the method

okuzan (changes requested):

okuzan (approved):
Good job

PR #860, opened 2023-12-18
implemented  method login and findByEmail

DmytroHadiuchko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
```
karma-o, line 6: this comment section is a javadoc of the `login()` method, which should be right above it, so let's do that, and the `userService` field along with the constructor field should be at the beginning of the class, in that exact order
[style guide](https://www.cs.cornell.edu/courses/JavaAndDS/JavaStyle.html#fielddecs)

karma-o (changes requested):

.github/workflows/ci.yml
```diff
@@ -1,18 +1,18 @@
...
-      - uses: actions/checkout@v2
-      - name: Set up JDK 17
-        uses: actions/setup-java@v2
-        with:
-          java-version: '17'
-          distribution: 'adopt'
-          cache: maven
-      - name: Build with Maven
-        run: mvn --batch-mode --update-snapshots verify
+name: Java CI
```
karma-o, line 1: It is weird that GitHub shows us that files were replaced with the same contents or changed names, but they did not, maybe you experimented with some branches or other things? Because it shouldn't be like that.
And it must have happened after my review, because it was Ok last time I saw your PR

karma-o (changes requested):
Strange files changes, we should fix that

karma-o (approved):

PR #861, opened 2023-12-18
solution added

NadVlados:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Rommelua, line 14: make it class field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
...
+        return userService.findByEmail(email) != null
+                && password.equals(userService.findByEmail(email).getPassword());
```
Rommelua, line 16: do not call findByEmail 2 times

Rommelua (changes requested):

Sanderfuck (approved):

PR #862, opened 2023-12-18
 Implemented the login method in AuthenticationService and the findBy…

GoldHolub:
…Email method in UserService.

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,11 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = null;
```
Sanderfuck, line 18: Why you don`t want to init user here?

GoldHolub: mere error)

Sanderfuck (approved):

PR #863, opened 2023-12-19
user service changes

margosalii:
added changes to UserServise&AuthenticationService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,6 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,14 @@ public class AuthenticationService {
...
+    private UserService userSer = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userSer.findByEmail(email);
+        if (user != null && password.equals(user.getPassword())) {
+            return true;
+        } else {
+            return false;
+        }
```
olekskov, line 24: There is no need in if/else when we need to return boolean value
```suggestion
        return user != null && password.equals(user.getPassword());
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (users[i].getEmail().equals(email)) {
+                return users[i];
+            }
+        }
```
olekskov, line 22: Please rewrite with for-each loop, it will increase code readability

for(User user : users) {...

olekskov (changes requested):

okuzan (approved):
✅

PR #865, opened 2023-12-21
Hm jv oop lesson 0

vladshender:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,29 @@
...
+        for (User user : users) {
+            if (user.getEmail().equals(email)) {
+                return user;
+            }
+        }
         return null;
     }
 }
+
+
```
okuzan, line 31: each file should end with exactly one empty line, you have two extra

okuzan (approved):
Good job!

PR #869, opened 2023-12-27
Added appropriate code

OleksandrStrilets:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,22 @@ public class AuthenticationService {
...
      */
+    private UserService currentuser = new UserService();
+
     public boolean login(String email, String password) {
+        if (currentuser.findByEmail(email) != null) {
+            if (currentuser.findByEmail(email).getEmail() == email
+                    && currentuser.findByEmail(email).getPassword() == password) {
+                return true;
+            }
+        }
```
Sanderfuck, line 22: At first save your user what you get from findByEmail method and reuse this and don`t compare String type by "=="

Sanderfuck (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (users[i].getEmail() == email) {
```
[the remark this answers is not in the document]
OleksandrStrilets: Ok clear, need to use '.equals' method

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (users[i].getEmail().equals(email)) {
+                return users[i];
+            }
+
```
Sanderfuck, line 22:
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,29 @@
...
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService currentuser = new UserService();
+
     public boolean login(String email, String password) {
+        if (currentuser.findByEmail(email) != null) {
+            User user = new User(currentuser.findByEmail(email).getEmail(),
+                    currentuser.findByEmail(email).getPassword());
```
Sanderfuck, line 21:
```suggestion
            User user = currentuser.findByEmail(email);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,29 @@
...
+            if (user.getPassword().equals(password) && user.getPassword().equals(password)) {
```
Sanderfuck, line 22: In this line you compare same in left and right side! Here you can check user for null and after this compare his password (you can write this in one line with return for simplifying code)

Sanderfuck: And after this you don`t need write "return true" and "return false"

Sanderfuck (changes requested):

Sanderfuck (changes requested):
Not fixed, check changes

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
...
      * Return false in any other cases.
      */
+    private UserService currentuser = new UserService();
+
     public boolean login(String email, String password) {
+        User user = currentuser.findByEmail(email);
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
         return false;
```
olekskov, line 23:
```suggestion
        return user != null && user.getPassword().equals(password);
```

olekskov (commented):
Good overall.
Further please do not use if/else or ternary when you need to return boolean value

olekskov (approved):

PR #874, opened 2023-12-29
implement a login method in AuthenticationService and findByEmail met…

balitskiy01:
…hod in UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -3,10 +3,7 @@ package mate.academy.service;
 import mate.academy.model.User;
 
 public class AuthenticationService {
-    private static final User[] users = new User[] {
-            new User("bob@i.ua", "1234"),
-            new User("alice@i.ua", "1234"),
-    };
+
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,29 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private static final User[] users = new User[] {
+            new User("bob@i.ua", "1234"),
+            new User("alice@i.ua", "1234"),
+    };
```
sokimaaa, line 9: Use UserService instead
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,29 @@
...
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        for (int i = 0; i < users.length; i++) {
+            if (users[i].getEmail().equals(email) && users[i].getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
```
sokimaaa, line 27: Dont complicate code with if statement. You can directly return the condition. 
```suggestion
        return /* condition */;
```

sokimaaa (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
sokimaaa, line 19: userService  should be instantiated once
consider making it class level field

```suggestion
        UserService userService = new UserService();
```

sokimaaa (changes requested):

PR #878, opened 2023-12-30
feat: implement login method

oksana-divkit:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,10 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private static final UserService userService = new UserService();
```
karma-o, line 6: Let's make it non-static, generally, it is easier to work with, and more OOP-like, which we will cover later on in the course :)

karma-o (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +16,8 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        return Arrays.stream(users)
```
karma-o, line 19: Just wanted to say that it is cool that you are familiar with streams, keep up the good work! :fire:

karma-o (approved):

PR #879, opened 2024-01-01
added solution of oop-lesson-0

yanna-stepanova:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
...
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User findUser = userService.findByEmail(email);
+        if (findUser != null) {
+            if (findUser.getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
```
okuzan, line 23: Rewrite this code without using true/false literals. You can directly return the result of the condition instead of using if/else or ternary operation to return true/false. For example:
  // Instead of this
  if (condition) {
      return true;
  } else {
      return false;
  }
  // You can do this
  return condition;

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
okuzan, line 16: make this service a private field instead of a local variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
...
+        User findUser = userService.findByEmail(email);
```
okuzan, line 17:
```suggestion
        User user = userService.findByEmail(email);
```

okuzan (changes requested):

okuzan (commented):
Good job overall!

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        return (user != null && (user.getPassword().equals(password))) ? true : false;
```
okuzan, line 19:
```suggestion
        return user != null && user.getPassword().equals(password));
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User item: users) {
```
```suggestion
        for (User item : users) {
```

okuzan (changes requested):

PR #886, opened 2024-01-02
implement methods login and findByEmail

isakov-github:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService service = new UserService();
+        User user = service.findByEmail(email);
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
         return false;
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
```
```suggestion
        for (User user : users) {
```
you can use IDEA autoformatter (Ctrl Alt L) when you're not sure

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
isakov-github: what's wrong with my implementaion? just for understanding..

src/main/java/mate/academy/service/UserService.java
isakov-github: maven didn't complain)

src/main/java/mate/academy/service/AuthenticationService.java
okuzan: nothing's technically wrong, it's a code style suggestion, when we deal with boolean expressions and need to return a boolean, there's no need to use `true` `false` literals - we can achieve the same with less code
// Instead of this
if (condition) {
    return true;
} else {
    return false;
}
// You can do this
return condition;

isakov-github: thx

okuzan (approved):
Well done

PR #891, opened 2024-01-04
login in AuthenticationService and findByEmail in UserService were im…

danmalins:
…plemented

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -11,6 +11,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (email.equals("bob@i.ua") && password.equals("1234")) {
+            return true;
+        }
+        if (email.equals("alice@i.ua") && password.equals("1234")) {
+            return true;
+        }
         return false;
     }
 }
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -10,11 +10,21 @@ public class UserService {
 
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public String findByEmail(String email) {
+        if (email.equals("bob@i.ua")) {
+            return email;
+        }
+        if (email.equals("alice@i.ua")) {
+            return email;
+        }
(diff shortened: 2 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (email.equals("bob@i.ua") && password.equals("1234")) {
+            return true;
+        }
+        if (email.equals("alice@i.ua") && password.equals("1234")) {
+            return true;
+        }
```
sokimaaa, line 19: Why two same if-statements? 
```suggestion
        if (email.equals("bob@i.ua") && password.equals("1234")) {
            return true;
        }
        if (email.equals("alice@i.ua") && password.equals("1234")) {
            return true;
        }
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (email.equals("bob@i.ua") && password.equals("1234")) {
+            return true;
+        }
+        if (email.equals("alice@i.ua") && password.equals("1234")) {
```
sokimaaa, line 17: you should get `User` and extract from object password and email, not to simply hardcode all users

src/main/java/mate/academy/service/UserService.java
```diff
@@ -10,11 +10,21 @@ public class UserService {
...
+    public String findByEmail(String email) {
+        if (email.equals("bob@i.ua")) {
+            return email;
+        }
+        if (email.equals("alice@i.ua")) {
+            return email;
+        }
+        if (email.equals("john@i.ua")) {
+            return email;
+        }
```
sokimaaa, line 27: code duplication, use loop-statement to reduce
```suggestion
        if (email.equals("bob@i.ua")) {
            return email;
        }
        if (email.equals("alice@i.ua")) {
            return email;
        }
        if (email.equals("john@i.ua")) {
            return email;
        }
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (email.equals("bob@i.ua") && password.equals("1234")) {
```
sokimaaa, line 14: why dont use UserService here?

sokimaaa (changes requested):

PR #896, opened 2024-01-05
initial pull request

bohdan2704:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        User foundUser = userService.findByEmail(email);
+        if (foundUser != null && foundUser.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
kerrrusha, line 22: try to inline it

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
bohdan2704: No problems, changed. But if statement looks a little bit readable

kerrrusha (approved):

PR #899, opened 2024-01-07
Login and find user services created

Recdt:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,14 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService usersData = new UserService();
```
karma-o, line 15: Let's call just this variable as it is - `userService`
Because it's not just the data, it also has a functionality that we can and should use :)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,14 @@ public class AuthenticationService {
...
+
     public boolean login(String email, String password) {
+        for (User user: usersData.getUsers()) {
+            if (user.getEmail().equals(email) && user.getPassword().equals(password)) {
+                return true;
+            }
+        }
```
karma-o, line 22: Let's use `findByEmail()` method from `userService` to not duplicate our logic and here we can just do the null check and password check
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@ public class UserService {
...
+            if (user.getEmail().equals(email)) {
+                return user;
+            }
+        }
         return null;
     }
+
+    public User[] getUsers() {
+        return users;
+    }
```
karma-o, line 28: And we can remove this method
```suggestion
```

karma-o (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,17 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        // I'm not sure what better, to call 2 times find by email
+        // methoud, or made an instance of find by email result?
+        if (userService.findByEmail(email) != null) {
+            if (userService.findByEmail(email).getPassword().equals(password)) {
```
karma-o, line 19: we should always strive for [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) - which stands for Don't Repeat Yourself, including situations like this, when we can put the result of this method in a variable to not call it 2 times 

Also, we can remove those `if` statements, combine their condition into a single one and return it's result right away :)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,17 @@ public class AuthenticationService {
...
+                return true;
+            }
+        }
+
```
karma-o, line 23: redundant empty line
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
...
      */
     public User findByEmail(String email) {
+        for (User user:users) {
+            if (user.getEmail().equals(email)) {
+                return user;
+            }
+        }
         return null;
     }
+
```
karma-o, line 25: same here

PR #901, opened 2024-01-07
Solution added

slizko1:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,8 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+import mate.academy.model.User;
+
+public class AuthenticationService extends User {
```
karma-o, line 5: Can we say that a service that checks the email & password IS A user?
User class represents some actual existing user with some info that we need to know about him. AuthenticationService should be treated as some office building for example, that has a mechanism to authenticate those users(`login()` method), so those are 2 different things and can not be mixed together because a building is not a person.

For now here are some links to read more about the [inheritance](https://www.geeksforgeeks.org/inheritance-in-java/), [composition vs inheritance](https://www.digitalocean.com/community/tutorials/composition-vs-inheritance), do not dig too deep, because we will cover those things in the course

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User newUser = new UserService().findByEmail(email);
```
karma-o, line 16: let's not create a new service each time we call a `login()` method, and make it an instance level variable of `AuthenticationService`
Also, this is basically not a new user, since we can say that he is already registered, so let's name this variable just `user`

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        if (newUser != null && newUser.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
karma-o, line 20: can be simplified a bit :)
```suggestion
        return newUser != null && newUser.getPassword().equals(password);
```

karma-o (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
andrii-hoienko: This comment is not resolved, you still create UserService each time when login methid is called

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+
+    private User user;
```
andrii-hoienko, line 7: What is the reason to make User global class variable? I suggest to make it local variable in login method

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
+
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     * @param email    - user's email
```
andrii-hoienko, line 13: What is the reason of this change?

andrii-hoienko (changes requested):
Some improvements reqired

src/main/java/mate/academy/service/AuthenticationService.java
slizko1: It was made by accident

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        return user != null && user.getPassword().equals(password);
     }
 }
+
```
karma-o, line 22: we should have only 1 empty line at the end of the file

karma-o (approved):
Good job!

PR #902, opened 2024-01-08
support login authentication

moskalenkosergii:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService useService = new UserService();
```
olekskov, line 16: Please initialize services you rely on in class variable not to create new object each time method is called

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
...
+        User user = useService.findByEmail(email);
+
+        if (user == null) {
+            return false;
+        }
+
+        if (!user.getPassword().equals(password)) {
+            return false;
+        }
+        return true;
```
olekskov, line 26: if structure usually not required when we aim to return boolean value
```suggestion
        return user != null && user.getPassword().equals(password);
```

olekskov (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user:users) {
```
```suggestion
        for (User user : users) {
```
make sure to format your file (Ctrl Alt L)

okuzan (approved):
good job

PR #903, opened 2024-01-09
Implemented a login method in AuthenticationService and findByEmail method in UserService.

ReamFOX:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -3,7 +3,11 @@
 import mate.academy.model.User;
 
 public class AuthenticationService {
-    User user;
+    private User user;
+
+    public AuthenticationService() {
+        this.user = null;
```
ReamFOX, line 9: initialize var user as null, because in other ways maven tell that its better make var local in method

karma-o: Yevhenii was talking about the `UserService` instance, user should be a local method var as it was before :)
Because services are usually used all over the place, and we do not need to create lots of instances of them since 1 is enough.
Next time if you have questions abut the comment feel free to tag a mentor and ask them in your personal chat on the platform

karma-o (changes requested):

karma-o (approved):
Nice!

PR #905, opened 2024-01-14
jv-oop-lesson-0 HW solution

luk14n:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
...
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null) {
+            String passwordFromDb = user.getPassword();
+            return passwordFromDb.equals(password);
+        }
         return false;
```
MentorMA, line 24: Simplify this to one line in return state

MentorMA (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,20 @@
...
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        return userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password);
```
okuzan, line 18: save the result of a service call to a local variable and use it, instead of calling service twice 🙏

okuzan (changes requested):

okuzan (approved):

PR #906, opened 2024-01-14
Fulfilled class AuthenticationService and class UserService

Yaromyros:
I did not touch class main as it was said as well as class User.

Let's start with AuthenticationService class. I created private variable object of UserService to get access to method "findByEmail" and getters of email and password. Then I fulfilled method login with checking if user email was found and if password of that user was same as it was given. If my statement was right it returned true and in others cases it returned false. In class UserService in method findByEmail, I checked every user in array of users and checked if given in method email equal to email that given in user object (in his email variable). Moreover, if its equivalent it returns user if not it returns null. 

In the end I checked code-style

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -10,7 +10,11 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private static final UserService userService = new UserService();
     public boolean login(String email, String password) {
+        if (userService.findByEmail(email) != null && userService.findByEmail(email).getPassword() == password) {
+            return true;
+        }
         return false;
     }
 }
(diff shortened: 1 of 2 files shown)


Yaromyros:
Last commit, I saw my problems. It was about checking if string of password or email is equals to their derivatives. So I used "==" instead of "equals()", which should be used for String variables comparison.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,13 @@
...
-    /**
-     * Imagine that some user wants to login to your site.
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
+    private static final UserService userService = new UserService();
```
Ivan95kos, line 4:
```suggestion
    private final UserService userService = new UserService();
```

Ivan95kos (approved):
Great job, with a minor comment

src/main/java/mate/academy/service/AuthenticationService.java
Yaromyros: I understand that we working with objects that is why we need to stay out of static? Or there is some wiser explanation? Thanks)

PR #907, opened 2024-01-15
add homework 2

EduarfF:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
karma-o, line 14: let's not create a new service each time we call a `login()` method, and make it an instance level variable of `AuthenticationService`

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@ public class AuthenticationService {
...
+        if (userService.findByEmail(email).getEmail() == email
+                && userService.findByEmail(email).getPassword() == password) {
```
karma-o, line 16: We can save the user that we found to a variable to avoid calling `findByEmail()` twice
We should also ensure that it is not null before we call getter methods to avoid a null pointer exception

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User it : users) {
```
karma-o, line 18: we should give a more meaningful names to our variables
```suggestion
        for (User user : users) {
```

karma-o (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
```
karma-o, line 7: This code section is a Javadoc of the login() method and should be right above it, class fields and constructors on the other hand should be at the beginning of the class

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,15 @@ public class AuthenticationService {
...
+    private UserService userService;
+
+    public AuthenticationService() {
+        this.userService = new UserService();
+    }
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        return user != null && user.getEmail().equals(email) && user.getPassword().equals(password);
```
karma-o, line 24: we can also say that email check is redundant here, because we have found that user by email

karma-o (approved):
GJ! Left a few minor comments to improve your solution

PR #910, opened 2024-01-17
implemented methods:

DanilKi:
login in AuthenticationService;
findByEmail in UserService.


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,10 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private static final UserService userService = new UserService();
```
okuzan, line 6:
```suggestion
    private final UserService userService = new UserService();
```

okuzan (approved):
Well done!

PR #913, opened 2024-01-19
Task jv-oop-lesson-0 completed

apgroupua:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +14,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Ivan95kos, line 17: `UserService` must be a field class
```suggestion
```

Ivan95kos (changes requested):
Good job overall!

src/main/java/mate/academy/service/AuthenticationService.java
apgroupua: Should I create a new pull request?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,10 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private static final UserService userService = new UserService();
```
kerrrusha, line 6: we want userService to be created individually for each new `AuthenticationService` object, not one for all
```suggestion
    private final UserService userService = new UserService();
```

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
apgroupua: Sorry, 
fixed ...!

kerrrusha (approved):

PR #915, opened 2024-01-21
Working with services

ghost:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,15 @@
...
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
+
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
okuzan, line 8: make user a private field instead of a local variable

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -3,13 +3,13 @@
...
     public boolean login(String email, String password) {
         UserService userService = new UserService();
-        // Find the user by email
-        User user = userService.findByEmail(email);
 
-        // Check if the user exists and the provided password matches the user's password
-        return user != null && user.getPassword().equals(password);
+        this.user = userService.findByEmail(email);
+
+        return this.user != null && this.user.getPassword().equals(password);
```
okuzan, line 13: check this out to know where we need `this`
https://stackoverflow.com/a/2411283
when we don't, no need to use it

okuzan (changes requested):
sorry, I meant a userService, user should be a local field :)

Ivan95kos (approved):
Nice

PR #917, opened 2024-01-23
Sollution for additional task course fundamentals

azzilian:
Updated findByEmail to return email
Updated login to make a check
Checkstyle check complete

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -11,6 +11,16 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userservice = new UserService();
+
+        if (userservice.findByEmail(email) == null) {
+            return false;
+        } else if ((email == userservice.findByEmail(email).getEmail())
+                && password == userservice.findByEmail(email).getPassword()) {
+            return true;
+        } else {
+            return false;
+        }
+
     }
 }
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
+            if (email.equals(user.getEmail())) {
+                return user;
+            }
+        }
         return null;
     }
 }


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,16 @@ public class AuthenticationService {
...
+        UserService userservice = new UserService();
+
+        if (userservice.findByEmail(email) == null) {
+            return false;
+        } else if ((email == userservice.findByEmail(email).getEmail())
+                && password == userservice.findByEmail(email).getPassword()) {
+            return true;
+        } else {
+            return false;
+        }
```
karma-o, line 23: Let's also reorganize our logic just a bit, because it is hard to read at the moment
to avoid complications in if & else, we can do something like this:
```java
if (// user not null
    // user email is fine
    /*user password is fine*/) {
    return true;
}
return false;
```

karma-o (changes requested):

azzilian:
New commits - Regretably in order to avoid call findbyemail 3 times I have to create variable in method - or I should create it out of method?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
```
karma-o, line 6: this comment section is a JavaDoc of the `login()` method which should be right above it

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,16 @@ public class AuthenticationService {
...
     public boolean login(String email, String password) {
-        return false;
+        User saveduser = userservice.findByEmail(email);
+        if ((saveduser != null) && (email.equals(saveduser.getEmail()))
+                && (password.equals(saveduser.getPassword()))) {
+            return true;
+        } else {
+            return false;
+        }
+
```
karma-o, line 25: we do not need so many braces, and now, we can optimize it even more by removing the `if` and `else` completely and returning the result of this condition right away since it returns exactly what we need to return in our method, like so:
```suggestion
        return saveduser != null && email.equals(saveduser.getEmail())
            && password.equals(saveduser.getPassword());
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,16 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userservice = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User saveduser = userservice.findByEmail(email);
```
karma-o, line 18: we are not doing any saving logic here, so I should name it just `user`
```suggestion
        User user = userservice.findByEmail(email);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,16 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userservice = new UserService();
```
karma-o, line 15: we should follow the lower camelCase naming convention for variables
```suggestion
    private UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
karma-o, line 18: yep, that the correct approach :+1: 
there is no way to remove the `user` variable from here and no need for that either, so it was just a misunderstanding, 
Next time feel free to tag the mentor and ask any questions regarding the review comments in personal chat on the platform :smiley:

karma-o (changes requested):
GJ! All that's left is some code formatting

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +14,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        return user != null && email.equals(user.getEmail())
```
andrii-hoienko, line 20: There is no reason to compare emain because email checking logic already defined in UserService#findByEmail, so we can just compare passwords

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,10 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private UserService userService = new UserService();
```
andrii-hoienko, line 6: Optional: Actually it is common practice to instantiate fields in constructor not in fields, such approach adds flexibility and, for example, allows you to replace a original class with its child class

andrii-hoienko (changes requested):
Some minor improvements required

karma-o (approved):
Nice!

PR #920, opened 2024-01-24
solved task

mykolazx3:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
karma-o, line 14: let's not create a new service each time we call a `login()` method, and make it an instance level variable of `AuthenticationService`

karma-o (changes requested):
GJ! Left a small comment to improve your solution

src/main/java/mate/academy/service/AuthenticationService.java
mykolazx3: > let's not create a new service each time we call a `login()` method, and make it an instance level variable of `AuthenticationService`

fixed

mykolazx3:
> GJ! Left a small comment to improve your solution

comment was added

mykolazx3:
??

karma-o:
> I saw you rejected my fixed task solution again, why??

It was not a mentor rejection, it was an automated CI check that failed because you made a mistake in the checkstyle, (see the screenshot)
image
Red cross there is clickable where you can see the details of why that happened

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,15 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
```
karma-o, line 12: This code section is a Javadoc of the `login()` method and should be right above it, class fields on the other hand should be at the beginning of the class

karma-o (approved):
GJ! Left a comment about the rejection that you have asked, and another minor thing

PR #921, opened 2024-01-24
add new constructor

mstefanchuk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +17,9 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
```
kozhukhovsky, line 22: u can realize method findByEmail in user service

kozhukhovsky (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
mstefanchuk: > u should realize method findByEmail in user service

Тобто клас AuthenticationService не потрібно чипати?

kozhukhovsky (approved):

PR #924, opened 2024-01-28
Update methods in AuthenticationService and fix missing spaces in UserService

dmarmul:
added description in commit


dmarmul:
check it pls

PR #926, opened 2024-01-29
implemented a 'login' and 'findByEmail' methods

kaiiiseeel:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,6 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+public class AuthenticationService extends UserService {
```
karma-o, line 3: It would be better to use composition over inheritance here, because we can not say that `AuthenticationService` IS A `UserService`, so let's make `UserService` a field of an `AuthenticationService`
more about it [here](https://www.digitalocean.com/community/tutorials/composition-vs-inheritance)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+
+        for (mate.academy.model.User user : users) {
```
karma-o, line 15: Let's import this class instead of using its full path here
```suggestion
        for (User user : users) {
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,7 +3,7 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
```
karma-o, line 6: let's revert this change and make it private again because we should follow the incapsulation principle from OOP(which we will cover later on in the course) and not give everybody access to our users, only via its public method `findByEmail()` which we should use in `AuthenticationService`

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+
```
karma-o, line 14: same here
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
...
+        for (mate.academy.model.User user : users) {
+            if (email.equals(user.getEmail())) {
+                if (password.equals(user.getPassword())) {
+                    return true;
+                }
+            }
+        }
```
karma-o, line 21: we do not need to iterate over users in this method, we already have `findByEmail()` method that we can use here, and then just compare passwords and that's it, just do not forget about the null check :)

karma-o (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
karma-o, line 14: Nice! Now we just need to make this variable a field(class level variable) of an `AuthenticationService` class to not create a new `UserService` each time we call `login()` method, but to reuse a single instance each time

karma-o (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
kaiiiseeel: If i understood right, i need to move the "UserService userService = new UserService();" out of the method to the class and make it private as well?

karma-o: Yep, next time if you have any questions do not be shy to tag me in your private chat on the platform and ask them there because it is hard to see it here :)

karma-o (approved):
Good job!

PR #927, opened 2024-01-29
implement a login method in AuthenticationService and findByEmail

HAIOVYI:
method in UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -11,6 +11,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != null && UserService.findByEmail(email).getPassword().equals(password)) {
+            return true;
+        }
         return false;
     }
 }
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
+        for (User user : users) {
+            if (user.getEmail().equals(email)) {
+                return user;
+            }
+        }
         return null;
     }
 }


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
karma-o, line 17: let's not change the method signature, make it non-static and try to make it work, to follow up with the OOP principles, which will be covered in details later on :)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != null
+                && UserService.findByEmail(email).getPassword().equals(password)) {
```
karma-o, line 15: We can save the user that we have found to a variable to avoid calling `findByEmail()` twice

karma-o (changes requested):

src/main/java/mate/academy/service/UserService.java
HAIOVYI: fixed

src/main/java/mate/academy/service/AuthenticationService.java
HAIOVYI: fixed

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,10 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private UserService us = new UserService();
```
andrii-hoienko, line 6:
```suggestion
    private UserService userService= new UserService();
```

andrii-hoienko (changes requested):
Good job, left minor comments

src/main/java/mate/academy/service/AuthenticationService.java
HAIOVYI: fixed

andrii-hoienko (approved):
Good job!

PR #929, opened 2024-01-31
First try of solving the task

YanHrdMkr:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
kerrrusha, line 17: this method should not be static. generally, when said 

> implement **method**

it's meant to implement not static method, because static are functions, not methods

```suggestion
    public User findByEmail(String email) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,7 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = UserService.findByEmail(email);
```
kerrrusha, line 16: create `userService` as `private final UserService userService` class-level field
```suggestion
        User user = userService.findByEmail(email);
```

kerrrusha (changes requested):

kerrrusha (approved):

PR #930, opened 2024-02-01
Implement Authentication functionality

DmitryLazyuka:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
okuzan, line 16: make userservice a private class-level field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        if (userService.findByEmail(email) == null) {
+            return false;
+        }
```
okuzan, line 19: call findByEmail only once in this method, it may be a costly operation, save the result of the call to a local variable
try to rewrite your method without using true/false literals

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        return user.getEmail().equals(email) && user.getPassword().equals(password);
```
okuzan, line 21: no need to check email again

okuzan (changes requested):
Good job overall

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User foundUser = userService.findByEmail(email);
+        if (foundUser == null) {
+            return false;
+        }
+        return foundUser.getPassword().equals(password);
```
olekskov, line 22: Usually when we need to return boolean we don't need if block or ternary (except complicated cases where we wish to split condition to be readable)
```suggestion
        return foundUser != null && foundUser.getPassword().equals(password);
```

olekskov (approved):
Good job

src/main/java/mate/academy/service/AuthenticationService.java
DmitryLazyuka: Thanks for the tip. I'll keep it in mind for the future

PR #933, opened 2024-02-01
Created Auth logic

TonyH277:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,25 @@
...
     public User findByEmail(String email) {
+        for (User user : users) {
+            if (user.getEmail() == email) {
+                return user;
+            }
+        }
         return null;
     }
 }
+
```
karma-o, line 27: redundant empty line
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,25 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
```
karma-o, line 6: I am almost sure that this declaration was according to the style convention, so please double-check this moment

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
...
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
karma-o, line 16: let's not create a new UserService each time, make it a class field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
...
+        if (userService.findByEmail(email) != null) {
+            if (userService.findByEmail(email).getPassword() == password) {
+                return true;
+            }
+        }
         return false;
```
karma-o, line 22: we can simplify it by removing the `if`s and returning the result right away by combining 2 conditions into 1 with `&&`

karma-o (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
TonyH277: do I need to use "else" or I could leave "return false", can not find proper explaining

src/main/java/mate/academy/service/UserService.java
karma-o: What I meant here was about the space between `[]` and `{`, but since our `users` array is effectively final and constant here - we can name it like you did, anyway this is not that important, so you can leave it as it is :)

karma-o (approved):
Approved with some comments for improvement

PR #935, opened 2024-02-02
Methods implementation

fedaktaras:
Implementations of two methods.

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        if (user.getEmail().equals(email) && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
     }
 }
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -3,6 +3,12 @@ package mate.academy.service;
 import mate.academy.model.User;
 
 public class AuthenticationService {
+    private final UserService userService;
+
+    public AuthenticationService() {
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
karma-o, line 16: Let's not create a new `UserService` each time we call login method, and move this var to a class-level

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        if (user.getEmail().equals(email) && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
karma-o, line 24: email is basically already checked, since we found user by it, so we can remove this check. Also, we can move all checks to a single condition and return the result of it right away without any `if` statements :wink:

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +17,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        try {
+            return Arrays.stream(users).filter(user -> user.getEmail().equals(email))
+              .findFirst().get();
+        } catch (NoSuchElementException exception) {
+            return null;
+        }
```
karma-o, line 25: It's very cool that you are familiar with streams already, but this is not a proper way to handle `NoSuchElementException`, first of all, it is a Runtime exception, which means that we should not catch it even if we can, and we should avoid the situation that produces it, so following our method logic we should return null in case we did not find a user, so we can use `orElse()` method of `Optional` like that
```suggestion
        return Arrays.stream(users)
            .filter(user -> user.getEmail().equals(email))
            .findFirst()
            .orElse(null);
```
there will be a whole lot more information about streams, optionals, exceptions and other related things later in the course, so I would recommend to avoid using things that we haven't covered yet, unless you are sure in what you are doing.
Expected solution here would be just using a for each loop and comparing emails

karma-o (changes requested):
Left some comments to the PR

fedaktaras:
I've committed to another branch.
My bad.
Fixed code according to code review.

andrii-hoienko (approved):
GJ!

PR #938, opened 2024-02-04
solve lesson-0 task

esidarapp:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
okuzan, line 16: make service a private field

okuzan (changes requested):

okuzan (approved):
Well done!

PR #939, opened 2024-02-04
implemented a login and findByEmail methods

vikkushnir:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,15 @@ public class AuthenticationService {
...
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
karma-o, line 24: we can just return the result of the `if` condition here :)
```suggestion
        return user != null && user.getPassword().equals(password);
```

karma-o (approved):
GJ! Left a minor comment to optimize your solution

src/main/java/mate/academy/service/AuthenticationService.java
vikkushnir: Understood, I even forgot about this option, need more practice, thank you

PR #943, opened 2024-02-07
Authen service

qreqit:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,18 @@
...
 public class AuthenticationService {
+    private UserService userService = new UserService();
+
+    public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
+        return false;
+    }
```
okuzan, line 14:
```suggestion
     return user != null && user.getPassword().equals(password);
    }
```
we can do it in one line

okuzan (approved):
good job

PR #946, opened 2024-02-09
delete comment

playeeer:


olekskov (changes requested):
Please recheck if you pushed all made changes.
I can see only removed comments now and no implementation of method at all

PR #947, opened 2024-02-10
changed methods findByEmail and login

Oleh-Voitykhivskyi:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
...
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+        if (user != null) {
+            if (user.getPassword().equals(password)) {
+                return true;
+            }
+        }
+
         return false;
```
olekskov, line 23: we don't need if/else construction to return boolean value
```suggestion
        return user != null && user.getPassword().equals(password);
```

olekskov (changes requested):

okuzan (approved):
Well done!

PR #949, opened 2024-02-13
hw is done

Kateryna-Takhtai:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private UserService userService;
```
kerrrusha, line 6: this will not work because you did not initialized this field, it will be null

kerrrusha (changes requested):

kerrrusha (approved):

PR #951, opened 2024-02-15

Alona-Martynova:
…ethod in UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService serviceInform = new UserService();
+        User user = new User(email, password);
+        boolean info = serviceInform.findByEmail(email)
+                != null && user.getPassword()
+                .equals(password) ? true : false;
+        return info;
     }
 }
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -3,18 +3,27 @@ package mate.academy.service;
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
             new User("bob@i.ua", "1234"),
             new User("alice@i.ua", "1234")
     };
 
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
      * @param email - the input parameter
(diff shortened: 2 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService serviceInform = new UserService();
```
okuzan, line 16: make it a private class-level field

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,27 @@
...
+            if (user.getEmail().equals(email)) {
+                return user;
+            }
+        }
         return null;
     }
 }
+
+
+
```
okuzan, line 29: this file has 4 empty lines. each file should have only one

src/main/java/mate/academy/service/AuthenticationService.java
okuzan, line 16:
```suggestion
        UserService userService = new UserService();
```
i am not saying that we should always name variables after their data types, but serviceInform is a bit unclear

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = new User(email, password);
```
okuzan, line 17: we should construct user from given fields, it's useless, just use those argument directly.
save result of call to the service to this user local variable instead

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        boolean info = serviceInform.findByEmail(email)
+                != null && user.getPassword()
+                .equals(password) ? true : false;
+        return info;
```
okuzan, line 21: rewrite your solution to not use boolean literals like true or false, use logical expressions directly
also - In situations like this, there is no need to store the object in a local variable if it will be immediately used. Instead, you can pass the expression directly without assigning it to a variable

okuzan (changes requested):
Good job so far

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,11 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        return user != null && user.getPassword()
+                .equals(password) ? true : false;
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

okuzan (changes requested):
`user.getPassword().equals(password)` - is a logical expression of its own, we don't need to wrap it with a ternary operator and use literals like true and false

okuzan (approved):
good job!

PR #954, opened 2024-02-17
the findByEmail method is implemented to search for a user with a val…

IDMFokin:
…id email address in the User Service class and the login method in the AuthenticationService class for user authentication

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User byEmail = userService.findByEmail(email);
+        if (byEmail != null) {
+            if (byEmail.getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
     }
+
 }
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
MentorMA, line 16: Don`t create object inside this method, every invoke of method will create new object but you need only one

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
...
+        User byEmail = userService.findByEmail(email);
```
MentorMA, line 17: byEmail - is not informative naming

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
...
+        if (byEmail != null) {
+            if (byEmail.getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
```
MentorMA, line 23: Try to simplify in one line after return

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
...
     }
+
```
MentorMA, line 25:
```suggestion
```

MentorMA: Don`t need empty line

MentorMA (changes requested):

MentorMA (changes requested):
See previous comments

PR #959, opened 2024-02-27
wrote the UserService method and the AuthenticationService method

maxkrup98:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
okuzan, line 16: make userService a private class-level variable to avoid recreating it each time this method is called

okuzan (changes requested):
GJ!

src/main/java/mate/academy/service/AuthenticationService.java
maxkrup98: I didn't understand specifically what needs to be done, as I cannot make the class userService private. Do you mean that I should add final?

maxkrup98: its zero-lesson task. еhe assignment specifies that i should write only two methods. i can try to rewrite, but I'm not sure if it's within the scope of the task.

PR #960, opened 2024-02-27
added auth and user service

Serhii-Nosulia:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +16,9 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        return Arrays.stream(users)
+                .filter(u -> u.getEmail().equals(email))
```
okuzan, line 20:
```suggestion
                .filter(user -> user.getEmail().equals(email))
```
avoid one-letter naming even in stream chains

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,9 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        User byEmail = userService.findByEmail(email);
```
okuzan, line 18: a bad name for a user variable
let's just make it a `user`
```suggestion
        User user = userService.findByEmail(email);
```

okuzan (changes requested):
Good job!

kozhukhovsky (approved):

PR #964, opened 2024-02-27
This method findByEmail searches for a user by their email address. I…

Ze1ld:
…f a user with this email address exists, the method returns their object. If there is no user with the specified email address, null is returned.

In the AuthenticationService class
This code implements a method for authenticating a user on a website. It checks whether the user with the specified email address exists and whether the entered password matches the user's password. If the user exists and the passwords match, true is returned; otherwise, false is returned.


okuzan (approved):
well done! 🚀

PR #966, opened 2024-02-27
fixed tests failure

sovisou:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
...
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
-    public boolean login(String email, String password) {
-        return false;
+    public boolean login(String password, String email) {
+        UserService userService = new UserService();
```
Rommelua, line 17: Make UserService a class field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
...
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        return user.getPassword().equals(password);
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

Rommelua (changes requested):

kerrrusha (approved):

PR #969, opened 2024-03-02
All test passed

VladSchweps:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,22 @@
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
olekskov, line 17: Services used should be class variables. There is no need to create new service for each login attempt

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,22 @@
...
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
```
olekskov, line 11: Please recheck if any IDE plugin or something reformats existing code. These changes look weird.
Pull request shouldn't contain changes it's not designed for

olekskov (changes requested):

PR #970, opened 2024-03-02
jv-oop-lesson-0

rus1980707:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
okuzan, line 21: make userService a class-level field, not a local variable
try to rewrite this implementation without using true/false explicitly, return boolean expressions directly

okuzan (changes requested):

okuzan (changes requested):
Good job! Let's refactor your solution a bit

PR #971, opened 2024-03-04
added solution

nikitaabo:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
...
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService service = new UserService();
+        User user = service.findByEmail(email);
+        if (user != null) {
+            return user.getPassword().equals(password);
+        } else {
+            return false;
+        }
```
okuzan, line 22: try to rewrite this without if-else and false literal

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService service = new UserService();
+        User user = service.findByEmail(email);
+        return user != null && user.getPassword().equals(password);
```
okuzan, line 18: nice!

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService service = new UserService();
```
okuzan, line 16: make userService a private field of the class, not a local variable

okuzan (changes requested):

Vasul-Pankiv (approved):

PR #972, opened 2024-03-04
add login validation and finding by email

VadikRakitin:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,17 @@ public class UserService {
...
+            if (users[i].getEmail().equals(email)) {
+                return users[i];
+            }
+        }
         return null;
     }
+
+    public static User[] getUsers() {
+        return users;
+    }
```
kerrrusha, line 29: this method is not required by the task. instead, you should use findByEmail method
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User[] users = UserService.getUsers();
+        for (int i = 0; i < users.length; i++) {
+            if (users[i].getEmail().equals(email) && users[i].getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
```
kerrrusha, line 22: use findByEmail method from UserService instead

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
VadikRakitin: In commit №2, I forgot to remove an unused getter getUsers(), so I made a third commit

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
kerrrusha, line 16: prev variant was correct, when userService was created once, as a class-level `private final` field
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
kerrrusha, line 21: this can be rewritten inline, think about how)

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        return (userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password));
```
Rommelua, line 17: do not call `findByEmail` method 2 times - it is not good for productivity

Rommelua (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        return (user != null && user.getPassword().equals(password));
```
kerrrusha, line 19: redundant parentheses
```suggestion
        return user != null && user.getPassword().equals(password);
```

kerrrusha (approved):

PR #973, opened 2024-03-04
Added verify solution

mhvaleria:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,18 @@ public class AuthenticationService {
...
     public boolean login(String email, String password) {
-        return false;
+        if (email.equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
     }
 }
+
```
okuzan, line 22: each file should end with one empty line (not shown on GitHub)
Here, you have two
Please pay attention to formatting 🙏

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,18 @@ public class AuthenticationService {
...
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        if (email.equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
okuzan, line 19: that's not what we should check
Create a private class-level field userService
here, call its method findByEmail to get the user
assert that the password is correct (return this logical expression)

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,26 @@
...
      */
     public User findByEmail(String email) {
+        for (User user : users) {
+            if (user.getEmail().equals(email)) {
+                return user;
+            }
+        }
+
         return null;
+
```
okuzan, line 26:
```suggestion
```

okuzan (changes requested):
Good job with your UserService
Now you need to fix Auth Service

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
...
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
         return false;
```
okuzan, line 24:
```suggestion
            return user != null && user.getPassword().equals(password);
```
let's rewrite this in one line

okuzan (approved):
Well done!

PR #975, opened 2024-03-04
Finished task

alisahoys:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService service = new UserService();
```
kerrrusha, line 16: userService should be created once, at the class scope, as a `private final` field
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = service.findByEmail(email);
+        if (user.getEmail().equals(password)) {
+            return true;
+        }
         return false;
```
kerrrusha, line 21: 1. what if user will be null?
2. try to rewrite inline

kerrrusha (changes requested):

PR #976, opened 2024-03-06
i hope this is correct solution of lesson 1

ShastkivRuslan:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
```
Rommelua, line 5: add UserService as class field
```suggestion
public class AuthenticationService {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = findByEmail(email);
+        if (user != null) {
+            if (user.getPassword().equals(password)) {
+                return user.getPassword().equals(password);
+            }
+        }
         return false;
```
Rommelua, line 23:
```suggestion
        return user != null && ...;
```

Rommelua (changes requested):

kerrrusha (approved):

PR #977, opened 2024-03-06
implement auth logic:

chertiav:
- login method in AuthenticationService
- findByEmail method in UserService

### DIFF
--- a/src/main/java/mate/academy/model/User.java
+++ b/src/main/java/mate/academy/model/User.java
@@ -1,8 +1,8 @@
 package mate.academy.model;
 
 public class User {
-    private String email;
-    private String password;
+    private final String email;
+    private final String password;
 
     public User(String email, String password) {
         this.email = email;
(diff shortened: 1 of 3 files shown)


src/main/java/mate/academy/model/User.java
```diff
@@ -1,8 +1,8 @@
 package mate.academy.model;
 
 public class User {
-    private String email;
-    private String password;
+    private final String email;
+    private final String password;
```
kerrrusha, line 5: revert this

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
```
kerrrusha, line 6: revert this
```suggestion
    private static final User[] users = new User[] {
```

kerrrusha (changes requested):

PR #984, opened 2024-03-12
Added authentication methods

igor031299:


src/main/java/mate/academy/Main.java
kerrrusha, line 1: revert this

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,11 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private static String passwordUser;
+    private static String emailUser;
+
```
kerrrusha, line 8: we dont need that
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +15,16 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final User[] userService = UserService.getUsers();
```
kerrrusha, line 18:
```suggestion
    private final UserService userService = ...;
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +15,16 @@ public class AuthenticationService {
...
+
     public boolean login(String email, String password) {
+        for (User user : userService) {
+            if (email.equals(user.getEmail()) && password.equals(user.getPassword())) {
+                return true;
+
+            }
+
+        }
         return false;
```
kerrrusha, line 28: userService.findByEmail method should be used here

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,23 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
+        return users;
+    }
+
```
kerrrusha, line 14: we dont need that
```suggestion
```

kerrrusha (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
igor031299: why we need to use exactly class UserService ? Without class User i can't use getters for email and password

kerrrusha (approved):

PR #985, opened 2024-03-12
done

vgladkyi:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,11 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        return user != null ? password.equals(user.getPassword())
+                : false;
```
```suggestion
        return user != null ? password.equals(user.getPassword());
```

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
vgladkyi: @okuzan error "':' expected" , when I removed 20th line.

olekskov (approved):

PR #988, opened 2024-03-15
oleksandr-sydorskyi jv-oop-lesson-0 15.03.2024

oleksandr-sydorskyi:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,17 @@ public class AuthenticationService {
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
olekskov, line 15: we do not need to create new UserService for each method call. Please move to class variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,17 @@ public class AuthenticationService {
...
+        if (userService.findByEmail(email) == null) {
+            return false;
+        }
+        return userService.findByEmail(email).getPassword().equals(password);
```
olekskov, line 19: Better to avoid double call of same method (findByEmail in this case) as our search will iterate twice through whole users array
```suggestion
        User user = userService.findByEmail(email);
        return user != null && user.getPassword().equals(password);
```

olekskov (changes requested):

kerrrusha (approved):

PR #989, opened 2024-03-17
fix methods findByEmail & login

VadymOziPantielieienko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,7 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = new UserService().findByEmail(email);
```
MentorMA, line 16: Create new user service in field, because you need only one object, every invoke of login method will create one more

MentorMA (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
VadymOziPantielieienko: New UserService field created

MentorMA (approved):

PR #990, opened 2024-03-17
Hometask completed

MarynaKushniruk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+
+        UserService userService = new UserService();
```
okuzan, line 15: make userService a class level field, to not recreate it on each method call

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,9 @@ public class AuthenticationService {
...
+        return userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password);
```
okuzan, line 17: don't call service/database twice, save result of the call

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+
```
```suggestion
```

okuzan (changes requested):

Rommelua (approved):

PR #993, opened 2024-03-20
Implemented findByEmail and login methods

DatkoAndrii:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,17 @@ public class AuthenticationService {
...
+    public AuthenticationService() {
+        this.userService = new UserService();
+    }
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
MentorMA, line 26: Let`s simplify this expression in one line after one return

MentorMA (changes requested):

Rommelua (approved):

PR #995, opened 2024-03-22
added solution of the task

AndriiLoto:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,16 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+    private boolean result = false;
```
okuzan, line 16: shouldn't be a class field, we don't need boolean variable at all here

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,16 @@ public class AuthenticationService {
...
+
     public boolean login(String email, String password) {
-        return false;
+        User myUser = userService.findByEmail(email);
+        if (myUser != null && password.equals(myUser.getPassword())) {
+            result = true;
+        } else {
+            result = false;
+        }
+        return result;
```
okuzan, line 25: (myUser != null && password.equals(myUser.getPassword()) is an expression that returns a boolean value, and you use it in `if` to return .. a boolean value. In cases like this, we can return the value directly, there's no need for `if`

okuzan (changes requested):
Good job! Let's simplify handling of booleans

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
MAMentorFX, line 16: You don't want to creat userService each time you call login method. Think, how can you create userService only once

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
...
+        User myUser = userService.findByEmail(email);
```
MAMentorFX, line 17:
```suggestion
        User user = userService.findByEmail(email);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,13 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        User myUser = null;
```
MAMentorFX, line 18: You don't need this. Just return user

MAMentorFX (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
```
MAMentorFX, line 15: You can make it final

MAMentorFX (approved):

PR #996, opened 2024-03-23
add findByEmail and login

playeeer:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
olekskov, line 16: Please make UserService class field. We do not need to create new object of service each time method is called

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
olekskov, line 22: When we need to return boolean value if/else structure or ternary is not required

instead of
if (condition) {
    return true;
} else {
    return false;
}

we may use just
return condition;

olekskov (changes requested):

playeeer (commented):
Delete User Service, create private UserService, delete if

PR #999, opened 2024-03-26
fixed checkstyle errors and added logic to methods

Jacalius:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
Rommelua, line 17: do not use static methods

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
Rommelua, line 20:
```suggestion
        return user != null && user.getPassword().equals(password));
```

Rommelua (changes requested):

src/main/java/mate/academy/service/UserService.java
Jacalius: If I don't use the static User method, I'll get an error during compilation (java: non-static method findByEmail(java.lang.String) cannot be referenced from a static context)

MentorMA: What approach do you need for invoke non-static method?

MentorMA: don`t make this method static

MentorMA (changes requested):

src/main/java/mate/academy/service/UserService.java
MentorMA: If you can`t resolve this problem, feel free ask in chat, mentors not answer in git

MentorMA (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService user1 = new UserService();
```
okuzan, line 16: make this service a class level field, use a better name, like userService, naming with numbers is a bad practice

okuzan (changes requested):
one issue left

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,14 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
+    public UserService getUserService() {
+        return userService;
+    }
+
```
okuzan, line 20: redundant

okuzan (approved):

src/main/java/mate/academy/service/AuthenticationService.java
Jacalius: removed redundant

PR #1004, opened 2024-03-29
implemented ligin and findByEmail methods

StVadym:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +11,9 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        return userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password);
```
olekskov, line 17: You call userService.findByEmail(email) twise, what means that service will perform search also twise.
Please introduce variable to have only one search call

olekskov (changes requested):

olekskov (approved):
Good job

PR #1005, opened 2024-03-30
make AuthenticationService login, UserService findByEmail methods work

HeroW1zard:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
okuzan, line 16: make this a private class-level field

okuzan (changes requested):
good job!

Rommelua (approved):

PR #1007, opened 2024-03-31
added authentication service

kriazantseva:


kriazantseva:
everything fine?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +14,13 @@ public class AuthenticationService {
...
      */
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
     }
+
```
MAMentorFX, line 25:
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+
```
MAMentorFX, line 18:
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
...
+        for (int i = 0; i < users.length; i++) {
+            if (users[i].getEmail().equals(email)) {
+                return users[i];
+            }
+        }
```
MAMentorFX, line 23: You can do like Map userEmailToPassword, to avoid O(n)

MAMentorFX (approved):

src/main/java/mate/academy/service/UserService.java
kriazantseva: I am confident that using a map can effectively optimize the code and speed up its compilation. However, I have not yet studied this in this course, so I cannot apply it. Most likely, this will be explained in the next topics, so do not consider my solution incorrect. I have removed the unnecessary lines you pointed out, as can be seen in the following commit

PR #1009, opened 2024-04-01
First commit.

maksym-bakunov:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,14 @@ public class AuthenticationService {
...
+
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        } else {
+            return user.getPassword().equals(password);
+        }
```
okuzan, line 23: try to rewrite this without if-else, and `false` literal, return boolean expression directly

okuzan (changes requested):

okuzan (changes requested):
.

Rommelua (approved):

PR #1010, opened 2024-04-02
Oop lesson 0 solution

kriazantseva:


kriazantseva:
At the previous pull request, the mentor mentioned "map", but I haven't learned anything about it yet. It must be somewhere in next lections.

PR #1018, opened 2024-04-06
Implemented searching by email

vovms:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User u: users) {
```
MentorMA, line 18:
```suggestion
        for (User user : users) {
```

MentorMA: Use informative variables

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService service = new UserService();
```
MentorMA, line 16: Create object in class scope, not in method

MentorMA: every invoke of method will create new object but you need only one object

MentorMA (changes requested):

MentorMA (changes requested):
Fix comments

src/main/java/mate/academy/service/UserService.java
vovms: Understood,
Than you for your time

kerrrusha (approved):

PR #1019, opened 2024-04-06
implemented fixes

ARiezin:


ARiezin:
my bad, probably, I modified a class main and it shouldn't be a right solution

ARiezin:
i'll solve and will do this properly

kerrrusha (changes requested):
cant see solution
image

ARiezin:
fixed, please review

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
         return false;
```
MentorMA, line 22: Simplify this using one return in one line

MentorMA (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,10 @@ public class AuthenticationService {
...
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        return user != null && user.getPassword().equals(password);
     }
 }
+
```
okuzan, line 22:
```suggestion
```

okuzan (approved):

PR #1021, opened 2024-04-09
implement a "login" and "findByEmail" methods

DukalisD:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
```
okuzan, line 20: userservice should be a class-level field
rewrite this solution without using `true` `false` boolean literals, return expression directly

okuzan (changes requested):

Rommelua (approved):

PR #1034, opened 2024-05-02
add task solution

snowfallen:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,28 @@
...
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
             new User("bob@i.ua", "1234"),
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
+        return users;
+    }
```
okuzan, line 13: remove this method. findByEmail is enough for our needs.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (User user : UserService.getUsers()) {
+            if (user.getEmail().equals(email) && user.getPassword().equals(password)) {
+                return true;
+            }
+        }
```
okuzan, line 21: make userService a class level variable
use findByEmail method
try to write your solution without using true and false literals, but with ternary operator

okuzan (changes requested):

kerrrusha (approved):

PR #1037, opened 2024-05-09
implemented oop lesson 0

ynik404:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private static final UserService AUTHENTICATION_SERVICE = new UserService();
```
okuzan, line 6:
```suggestion
    private final UserService userService = new UserService();
```
it's not a constant, and shouldn't be one - we just inject this dependency

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
...
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        User userFromDb = AUTHENTICATION_SERVICE.findByEmail(email);
+        if (userFromDb == null) {
+            return false;
+        }
+        return userFromDb.getPassword().equals(password);
```
okuzan, line 24: try to rewrite it in one line, without use false literal

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,20 @@
...
      * @param password - user's password
+     *                 new User("bob@i.ua", "1234"),
+     *             new User("alice@i.ua", "1234")
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        return userService.findByEmail(email).getPassword().equals(password);
```
okuzan, line 18: this can throw NPE -- you should still check for null :)  You should have one line to get user from service, and one line to return an expression (use && there)

okuzan (changes requested):

MentorMA (approved):

PR #1039, opened 2024-05-14
solved task jv-oop-lesson-0

ASelivonchyk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        try {
+            return service.findByEmail(email).getPassword().equals(password);
+        } catch (RuntimeException e) {
+            return false;
+        }
```
kerrrusha, line 20: Runtime Exceptions are not meant to be handled
```suggestion
        return service.findByEmail(email).getPassword().equals(password);
```

koval-kn (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
ASelivonchyk: edited.

PR #1041, opened 2024-05-15
Added solution of homework

aovozniuk1:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (users[i].getEmail().equals(email)) {
+                return users[i];
+            }
+        }
```
MentorMA, line 22: Better use for each loop

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +11,14 @@ public class AuthenticationService {
...
      */
+
     public boolean login(String email, String password) {
+        if (userService.findByEmail(email) == null) {
+            return false;
+        }
+        if (userService.findByEmail(email).getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
MentorMA, line 22: Save user for variable that you get from `userService.findByEmail(email)` and after this check for password

MentorMA (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,10 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private UserService userService = new UserService();
+    private User user;
```
kerrrusha, line 7: should be method-level
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +14,12 @@ public class AuthenticationService {
...
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        return user.getPassword().equals(password);
```
kerrrusha, line 23: think about how to rewrite it inline

koval-kn (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
aovozniuk1: Sorry, I didn't get what should l do. What does mean inline?

MentorMA: ` if (user == null) {
            return false;
        }
        return user.getPassword().equals(password);`

it`s mean that you need write this code

in one line after return, because expression return you boolean value, so that you don`t need to write return true or return false

MentorMA: if you have question ask it in chat

MentorMA (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
aovozniuk1: But now check style fails because of this changes Screenshot_23

okuzan (changes requested):
build failed. Run smth like `mvn clean package` to check if build passes locally before submitting your solution 🤝

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,9 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        return user == null ? false : user.getPassword().equals(password) ? true : false;
```
MentorMA, line 19:
```suggestion
        return user != null && user.getPassword().equals(password);
```

MentorMA: Simplify

MentorMA (changes requested):

Elena-Bruyako (approved):

PR #1043, opened 2024-05-23
hw-jv-oop-lesson-0

ihor-sydorenko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,20 @@
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        return email.equals(userService.findByEmail(email).getEmail())
```
Elena-Bruyako, line 17: It's better to create User local variable
`User user = userService.findByEmail(email);`
and then check if user != null and have equals password 

In such approach, you won't call userService.findByEmail(email) twice and won't get NullPointerException if User is null

Elena-Bruyako (changes requested):

MentorMA (approved):

PR #1044, opened 2024-05-23
implemented solution

maaarynet:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != null && User.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
Vasul-Pankiv, line 19:
```suggestion
        return UserService.findByEmail(email) != null && User.getPassword().equals(password);
```

Vasul-Pankiv (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
kerrrusha, line 17: it should not be static. 
```suggestion
    public User findByEmail(String email) {
```

kerrrusha (changes requested):

okuzan (approved):
well done

PR #1045, opened 2024-05-23

Pancher07:
…ethod in UserService

### DIFF
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -3,18 +3,27 @@ package mate.academy.service;
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,27 @@
...
     public User findByEmail(String email) {
+        if (email == null || email.isEmpty()) {
+            return null;
+        }
+        for (User user : users) {
+            if (email.equals(user.getEmail())) {
+                return user;
+            }
+        }
         return null;
```
Vasul-Pankiv, line 27:
```suggestion
        for (User user : users) {
            if email  == user.getEmail() || email != null && email .equals(user.getEmail()) {
                return user;
            }
        }
        return null;
```

Vasul-Pankiv (changes requested):

Elena-Bruyako (approved):

PR #1046, opened 2024-05-26
added implementation of login and findByMail methods

Tetiana-Hrytsenko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,7 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = new UserService().findByEmail(email);
```
Elena-Bruyako, line 16: For the future, it's better to:
```
```

Elena-Bruyako (approved):

PR #1049, opened 2024-06-06
Solution

vladis9:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Vasul-Pankiv, line 16: Make it class level field to avoid creating new object per each method call

Vasul-Pankiv (changes requested):

Elena-Bruyako (approved):

PR #1052, opened 2024-06-10
make a authentication metod

deez-nerdy:
make a authentication metod & correct sheckstyle mistakes

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        if (userService.findByEmail(email) != null) {
```
Elena-Bruyako, line 16: How can you avoid calling findByEmail method twice here?

Elena-Bruyako (changes requested):

Elena-Bruyako (changes requested):
See the comment

src/main/java/mate/academy/service/AuthenticationService.java
deez-nerdy: i try to avoid it by saving user in login method. i make new pull request

okuzan (approved):
well done

PR #1054, opened 2024-06-16
implemented the login method in AuthenticationService

Dimon1888:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        String qw = "1234";
```
Elena-Bruyako, line 14: You should create UserService object, call findByEmail method and return true if user by email exists and passed password is equal to user's password.
Return false in any other cases.

Elena-Bruyako (changes requested):

PR #1058, opened 2024-06-30
Created method login

Andormy:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,21 @@ public class AuthenticationService {
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        if (email.equals(email) && password.equals(password)) {
```
Elena-Bruyako, line 15: You compare that email the same as email, same for password. 
Read task description, read comment for methods and implement the logic. 
Is you have some questions, ask in chat

Your task is to implement a login method in AuthenticationService and findByEmail method in UserService

Elena-Bruyako (changes requested):

PR #1059, opened 2024-06-30
added methods implementation

ykudrynska:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
okuzan, line 14: make this service a private class field to avoid redundant recreation per each method invocation

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
...
+        if (userService.findByEmail(email) != null) {
+            if (userService.findByEmail(email).getPassword().equals(password)) {
+                return true;
+            } else {
+                return false;
+            }
+        }
         return false;
```
okuzan, line 22: try to rewrite this without using true/false literals, only ternary operator

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
ykudrynska: Thanks, fixed

src/main/java/mate/academy/service/AuthenticationService.java
ykudrynska: Fixed. Applied there ternary operator, but Intellij IDEA suggested to simplify the expression even more, applied suggestion.

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        if (userService.findByEmail(email) != null) {
+            return userService.findByEmail(email).getPassword().equals(password);
+        }
         return false;
```
MentorMateAcademy, line 19: you duplicate `userService.findByEmail(email)` let save it to variable and write it using one return

MentorMateAcademy (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
ykudrynska: @MentorMateAcademy, fixed, thanks for pointing on that!

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null) {
```
Elena-Bruyako, line 19: could you make one line of return instead of 4?

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
ykudrynska: @Elena-Bruyako , cool) thanks! fixed

Elena-Bruyako (approved):

PR #1068, opened 2024-07-23
Fix AuthenticationService and UserService implementation

SIPervii:
- Implement findByEmail method in UserService to search for a user by email from the static array of users.
- Correct the login method in AuthenticationService to use UserService for email lookup and validate the provided password.
- Remove duplicate login method in AuthenticationService.
- Ensure correct logic for user authentication based on email and password.

This fix ensures the correct functioning of the login process, and the main method now passes all test cases as expected.


MentorMateAcademy (changes requested):
Where is implemetnation of User Service?

SIPervii:
Чому другий раз відхилили?

PR #1070, opened 2024-07-24
AuthenticationService login method and findByEmail method are impleme…

DmDvd:
…nted

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,6 +1,14 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private UserService userService;
+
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +19,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
Elena-Bruyako, line 26:
```suggestion
        return user != null && user.getPassword().equals(password);
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1071, opened 2024-07-29

FireFalcons:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,20 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
```
MentorMateAcademy, line 5:
```suggestion
public class AuthenticationService {
```

MentorMateAcademy: it`s not good hierarchy inheritance

MentorMateAcademy (changes requested):

PR #1073, opened 2024-08-01
added solution

aunkerrr:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userServiece = new UserService();
+        User user = userServiece.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
okuzan, line 21: try to rewrite this without using true and false literals, just return boolean expression
user service should be a private class-level field, not a local one

okuzan (changes requested):

PR #1074, opened 2024-08-02
implemented AuthenticationService and UserService

rainett:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +16,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        Optional<User> userOptional = Optional.ofNullable(userService.findByEmail(email));
+        return userOptional.map(user -> user
+                .getPassword()
+                .equals(password)).orElse(false);
```
okuzan, line 22: very nice!

PR #1075, opened 2024-08-02
Implemented simple authentication

NataBurtnyk:


okuzan (changes requested):
Build failed

NataBurtnyk:
Fixed

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
okuzan, line 16: let's make UserService a private class-level field, to avoid needless recreation per each method invocation

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+
+        if (user != null) {
+            if (user.getPassword().equals(password)) {
+                return true;
+            }
+        }
+
         return false;
```
okuzan, line 25: try to rewrite this solution without using true/false literals, just operation with boolean expression, return it directly

okuzan (approved):

src/main/java/mate/academy/service/AuthenticationService.java
NataBurtnyk: Fixed remarks

okuzan (approved):

PR #1077, opened 2024-08-04
completed jv_lesson_0

Oleg021:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        if (userService.findByEmail(email) != null) {
```
Elena-Bruyako, line 17: Don't call userService.findByEmail(email) twice, create local variable
also, you can use ternary operator

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,20 @@ public class AuthenticationService {
...
+    public void setUserService(UserService userService) {
+        this.userService = userService;
+    }
+
     public boolean login(String email, String password) {
-        return false;
+        return userService.findByEmail(email) != null && userService
+                .findByEmail(email)
+                .getPassword()
+                .equals(password) ? true : false;
```
```suggestion
        User user = userService.findByEmail(email);
        return user != null && user.getPassword().equals(password);
```

Elena-Bruyako (changes requested):

PR #1078, opened 2024-08-04
implement login() method in AuthenticationService;

mkarmadev:
implement findByEmail() method in UserService;

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,17 @@ public class AuthenticationService {
...
+
+        if (user == null) {
+            return false;
+        }
+
+        if (password.equals(user.getPassword())) {
+            return true;
+        }
+
         return false;
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1080, opened 2024-08-16
les_0 done

VolodymyrNyshta:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userServ = new UserService();
```
```suggestion
        UserService userService = new UserService();
```

should be class-level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = userServ.findByEmail(email);
+        if (user != null && user.getPassword() == password) {
+            return true;
+        }
         return false;
```
Elena-Bruyako, line 21:
```suggestion
        return user != null && user.getPassword() == password;
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,13 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        if (email != null) {
+            for (int i = 0; i < users.length; i++) {
+                if (users[i].getEmail() == email) {
+                    return users[i];
+                }
+            }
+        }
```
Elena-Bruyako, line 24:
```suggestion
        for (User user : users) {
            if (user.getEmail().equals(email)) {
                return user;
            }
        }
```

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
VolodymyrNyshta: ok. Can I fix it here or create new commit and pull request?

Elena-Bruyako (approved):

PR #1081, opened 2024-08-20

skotsurskiy:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = new User(email, password);
+        if (new UserService().findByEmail(email) != null
```
okuzan, line 17: UserService should be a private class-level field, not an anonymous instance

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+                && user.getPassword().equals(new UserService().findByEmail(email).getPassword())) {
```
okuzan, line 18: moreover, you create it twice per each method call. You should create one per each method _instance_

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+            return true;
+        }
         return false;
```
okuzan, line 21: try to rewrite your solution without using true/false boolean literals 😉

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,11 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = new User(email, password);
```
```suggestion
```
use values directly

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,11 @@ public class AuthenticationService {
...
+        User checkuser = userService.findByEmail(email);
```
okuzan, line 19:
```suggestion
        User user= userService.findByEmail(email);
```

okuzan (changes requested):
almost done

Elena-Bruyako (approved):

PR #1082, opened 2024-08-23
Added logic to findByEmail and login methods

Petliuk:


Elena-Bruyako (changes requested):
build failed, fix it

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
      * Return false in any other cases.
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null && password.equals(user.getPassword())) {
+            return true;
+        }
         return false;
```
MentorMateAcademy, line 23: write it one line using one return

MentorMateAcademy (changes requested):

Elena-Bruyako (approved):

PR #1084, opened 2024-09-02
changes in login

Anasteisha4:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,39 @@
...
      */
     public boolean login(String email, String password) {
+        if (email.equals(validEmailFirst) && password.equals(validPasswordFirst)) {
+            return true;
+        }
+        if (email.equals(validEmailSecond) && password.equals(validPasswordFirst)) {
+            return true;
+        }
         return false;
     }
```
Elena-Bruyako, line 38:
```suggestion
    public boolean login(String email, String password) {
        UserService userService = new UserService();
        User user = userService.findByEmail(email);
        return user != null && user.getPassword().equals(password);
    }
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,39 @@
...
+        return validEmailFirst;
+    }
+
+    public String getValidEmailSecond() {
+        return validEmailSecond;
+    }
+
+    public String getValidPasswordFirst() {
+        return validPasswordFirst;
+    }
```
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,39 @@
 package mate.academy.service;
 
 public class AuthenticationService {
```
Elena-Bruyako, line 3: You should also implement UserService

Elena-Bruyako (changes requested):

PR #1085, opened 2024-09-03
initialization check

VovaPryt:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,14 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
+
     public User findByEmail(String email) {
+
```
okuzan, line 19:
```suggestion
```
dont' start a method/class with an empty line

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,17 @@ public class AuthenticationService {
...
+        User byEmail = userService.findByEmail(email);
+        User user = byEmail;
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
+
         return false;
     }
 }
+
```
okuzan, line 29:
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,17 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+
+        UserService userService = new UserService();
```
okuzan, line 19: You create new user service instance each time method is called, it's inefficient, let's create on per class, move it to a class level constant, private final

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,17 @@ public class AuthenticationService {
...
+        User byEmail = userService.findByEmail(email);
+        User user = byEmail;
```
```suggestion
        User user = userService.findByEmail(email);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,17 @@ public class AuthenticationService {
...
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
+
         return false;
```
okuzan, line 26: try to rewrite this solution without using true/false literals 😉

okuzan (changes requested):

okuzan (approved):

PR #1087, opened 2024-09-06
add solution

o-bihun:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako, line 16: userservice should be class-level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        return user != null && (user.getPassword().equals(password));
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

Elena-Bruyako (changes requested):

PR #1090, opened 2024-09-06
added solution

o-bihun:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako, line 16: it should be class-level variable
don't need to create new UserService() each time you call login method

Elena-Bruyako (changes requested):

okuzan (approved):

PR #1091, opened 2024-09-06
added solution

sergiikonov:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = new UserService().findByEmail(email);
```
okuzan, line 18: dont' create new UserService each time method is called, make it a private class level field

okuzan (changes requested):
almost done

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,9 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        User user = new UserService().findByEmail(email);
```
```suggestion
        User user = userService.findByEmail(email);
```

when you use `new UserService()` - you create new object, you already have it (class-level variable)

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1095, opened 2024-09-12
Homework

OlehRidchenko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,13 @@ public class AuthenticationService {
...
      */
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
```suggestion
        }
        return false;
```

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
okuzan, line 22: try to rewrite this with a ternary operator

okuzan (approved):

PR #1098, opened 2024-09-13
added solution for h/w-1

Nazar1106:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,27 @@
...
+
     public boolean login(String email, String password) {
-        return false;
+
+        User user = userService.findByEmail(email);
+
+        if (user == null) {
+            return false;
+        }
+        return user.getEmail().equals(email) && user.getPassword().equals(password);
```
Elena-Bruyako, line 25:
```suggestion
        User user = userService.findByEmail(email);
        return user != null && user.getPassword().equals(password);
```

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
Nazar1106: Are there check style mistakes in code?

okuzan (approved):

src/main/java/mate/academy/service/AuthenticationService.java
okuzan: looks good

PR #1099, opened 2024-09-14
Done all task

htsarenko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,18 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private static final UserService userService = new UserService();
```
okuzan, line 13:
```suggestion
    private final UserService userService = new UserService();
```
don't overuse static, let's just keep it final, it will become clear why when Spring Boot section starts

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,18 @@ public class AuthenticationService {
...
+
     public boolean login(String email, String password) {
+        if (email == null || password == null) {
+            return false;
+        }
+
```
okuzan, line 19: validation is most probably done on previous stages, assume it's out of scope of the task
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,18 @@ public class AuthenticationService {
...
+        if (userService.findByEmail(email) != null
+                && password.equals(userService.findByEmail(email).getPassword())) {
+            return true;
+        }
+
         return false;
```
okuzan, line 25: - don't call find by email twice, save result to a variable
- don't use true/false literals, return expression itself

okuzan (changes requested):

okuzan (approved):

PR #1101, opened 2024-09-18
Implement login & findByEmail methods

vDmytriv01:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = new UserService().findByEmail(email);
```
MentorMateAcademy, line 17: UserService should be class level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
+        if (user != null && user.getEmail().equals(email) && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
MentorMateAcademy, line 21: try to use one return and condition

MentorMateAcademy (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
      * You should check if user credentials (login and password) are valid or not.
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService userService = new UserService();
```
JJJazl, line 16: 1) You have to declare variable before java doc for method
2) It's better to use a constructor

JJJazl (changes requested):

okuzan (approved):

okuzan (approved):

PR #1105, opened 2024-09-24
complited

gubber230:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako, line 14: don't create UserService userService object each time when login method is called, create class-level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
...
+        return userService.findByEmail(email) != null
+                && userService.findByEmail(email).getPassword().equals(password);
```
```suggestion
        User user = userService.findByEmail(email);
        return user != null
                && user.getPassword().equals(password);
```

don't call method twice, you can create local variable

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private static final UserService userService = new UserService();
```
Elena-Bruyako, line 6:
```suggestion
    private final UserService userService = new UserService();
```

Elena-Bruyako (changes requested):

sarakhmen (approved):
lgtm!

PR #1106, opened 2024-09-25
implemented method findByEmail and method login

Cergei998:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,9 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        User user = UserService.findByEmail(email);
```
Elena-Bruyako, line 17: create an instanse of UserService 
it should be class-level variabel

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,13 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+
+    public static User findByEmail(String email) {
```
Elena-Bruyako, line 18:
```suggestion
    public User findByEmail(String email) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako, line 17: in previous comment, I mentioned that UserService should be class-level variable
Now, you create new instanse of UserService each time when you call login method

Elena-Bruyako (changes requested):

ahoienko (approved):
Good job!

PR #1107, opened 2024-09-25
add login method in AuthenticationService and findByEmail method in U…

jekaGr:
…serService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -3,6 +3,8 @@ package mate.academy.service;
 import mate.academy.model.User;
 
 public class AuthenticationService {
+    private UserService userService = new UserService();
+
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = new User(email,password);
+        UserService userService = new UserService();
```
ahoienko, line 17: This should be class level field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
...
+
+        if (userService.findByEmail(user.getEmail()) != null
+                && userService.findByEmail(user.getEmail()).getPassword().equals(user
+                .getPassword())) {
```
ahoienko, line 21: Code duplication (userService.findByEmail)

ahoienko (changes requested):
Very nice) Lets improve your solution a bit

PR #1110, opened 2024-09-30
add login and findByEmail method

Bohdan-Snute:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
MentorMateAcademy, line 17:
```suggestion
    public User findByEmail(String email) {
```

MentorMateAcademy: you don`t need static here

MentorMateAcademy (changes requested):

PR #1111, opened 2024-09-30
Added solution to the task. Modified files: AuthenticationService and…

nikolasdymchenko7:
Added solution to the task. 
Modified files: AuthenticationService and UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
MentorMateAcademy, line 16: create object in class level

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,15 @@ public class AuthenticationService {
...
+        User userByEmail = userService.findByEmail(email);
+
+        if (userByEmail != null) {
+            if (userByEmail.getEmail().equals(email)
+                    && userByEmail.getPassword().equals(password)) {
+                return true;
```
MentorMateAcademy, line 22: you don`t need to check email cause you get this user by email already

MentorMateAcademy (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
nikolasdymchenko7: make sense!

Elena-Bruyako (approved):

PR #1112, opened 2024-10-03
new push

Volikk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako, line 16: it should be class-level variable

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
Volikk: I need deleted this line?

PR #1117, opened 2024-10-09
Develop

oKukHtina:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private final UserService userService = new UserService();
```
JJJazl, line 6: It's better to use a constructor

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,14 @@ public class AuthenticationService {
...
      */
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
+
         return false;
```
JJJazl, line 24: You can simplify it by making one return statement:
`return user != null && user.getPassword().equals(password);`

JJJazl (changes requested):

PR #1118, opened 2024-10-16
Implementation of solution for jv-oop-lesson-0

SnkvS:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +15,11 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
+
     public User findByEmail(String email) {
-        return null;
+        return Arrays.stream(users)
```
Elena-Bruyako, line 20: stream works slower than for each, so for this particular task you can use for each loop

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
```
Elena-Bruyako, line 19:
```suggestion
        return user != null && user.getPassword().equals(password);
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1121, opened 2024-10-21
Initial commit for hw-solution

katrienkraska:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
Sanderfuck, line 22: Just write this in one line and use one return

Sanderfuck (changes requested):

okuzan (approved):

PR #1122, opened 2024-10-23
виходить ошибка для джона ,і не розумію в чому проблема.

mixx1k:


Elena-Bruyako (changes requested):
build failed, fix it first

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
```
```suggestion
public class AuthenticationService {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = findByEmail(email);
+        if (user != null) {
```
Elena-Bruyako, line 18: it's better to use ternary operator here

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
...
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = findByEmail(email);
```
Elena-Bruyako, line 17: you should create UserService instance. Think should it be class-level or local variable

Elena-Bruyako (changes requested):
If you have some questions, please ask in chat

PR #1126, opened 2024-10-29

Danila2006:
…ethod in UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -17,6 +17,7 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
         User user = userService.findByEmail(email);
         return user != null && password.equals(user.getPassword());


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Elena-Bruyako, line 16: it should be class-level variable, don't need to create `new UserService()`each time you call `login` method

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
```
Elena-Bruyako, line 18: minor comment - you can use for each loop here

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user != null && password.equals(user.getPassword())) {
```
Elena-Bruyako, line 18: minor comment - you can use ternary operator here

Elena-Bruyako (changes requested):

ahoienko (approved):
Good job!

PR #1127, opened 2024-10-30

olieniev:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService service = new UserService();
```
Elena-Bruyako, line 14: It should be class-level variable, don't need to create `new UserService()` object each time you call `login` method

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
...
+        return service.findByEmail(email) != null
```
Elena-Bruyako, line 15: Can you avoid calling findByEmail twice? 

Let's create local variable for the result of this metho

src/main/java/mate/academy/service/AuthenticationService.java
Elena-Bruyako, line 14:
```suggestion
        UserService userService = new UserService();
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1128, opened 2024-10-30
AuthenticationService is implemented

VitaliyJvM:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,10 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private static final UserService userService = new UserService();
```
Elena-Bruyako, line 6:
```suggestion
    private final UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +15,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user == null) {
```
Elena-Bruyako, line 19: you can use ternary operator

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1131, opened 2024-11-05
solved exercise

opolskyy:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+
+        UserService userService = new UserService();
```
ahoienko, line 15: What is the reason to create this object each time method login is called?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@ public class AuthenticationService {
...
+        if (userService.findByEmail(email) == null) {
+            return false;
+        }
+        return password.equals(userService.findByEmail(email).getPassword());
```
ahoienko, line 19: This expression is overcomplicated. I suggest you to make this operation in 2 lines to make code more readable

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,13 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+
+        for (int i = 0; i < users.length; i++) {
+
```
ahoienko, line 20:
```suggestion
```

ahoienko (changes requested):
Nice, left some comments

src/main/java/mate/academy/service/AuthenticationService.java
opolskyy: How else I could call method findByEmail if I wouldnt create this object?

sarakhmen (approved):

PR #1132, opened 2024-11-11
login method in AuthenticationService and findByEmail method in UserS…

kimanko:
…ervice were implemented

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -12,11 +12,13 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        User user = new UserService().findByEmail(email);
+        User user = userService.findByEmail(email);
         if (user == null) {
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = new UserService().findByEmail(email);
```
liliia-ponomarenko, line 16: `private UserService userService = new UserService(); `
make it class level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
...
+        if (user == null) {
+            return false;
+        }
+        return user.getEmail().equals(email) && user.getPassword().equals(password);
```
```suggestion
        return user.getPassword().equals(password);
```

liliia-ponomarenko (changes requested):
Good job! Let’s fix a few mistakes ;)

kimanko:
made UserService class level variable, fixed logical mistake

liliia-ponomarenko (approved):
Great!

PR #1136, opened 2024-11-16
Fixed project

DmitryShmerchuk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = new UserService().findByEmail(email);
```
okuzan, line 16: create userService once per class instance - make it a private class level field

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
...
+        if (user == null) {
+            return false;
+        }
+        return user.getPassword().equals(password);
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

okuzan (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
DmitryShmerchuk: Ready

DmitryShmerchuk:
Check again. I think I fixed everything

DmitryShmerchuk (commented):
Check again. I think I fixed everything

src/main/java/mate/academy/service/AuthenticationService.java
okuzan: not fixed, not a big deal though

okuzan (approved):

PR #1137, opened 2024-11-22

LiudmylaKuzmenko:
…ethod in UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
@@ -10,7 +12,20 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
+    private static final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+
+        if (user == null) {
+            return false;
+        }
+
+        if (password.equals(user.getPassword())) {
+            return true;
+        }
+
         return false;
     }
 }
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
+            if (email.equals(user.getEmail())) {
+                return user;
+            }
+        }
+
         return null;
     }
 }


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,20 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
+    private static final UserService userService = new UserService();
```
```suggestion
    private final UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,20 @@ public class AuthenticationService {
...
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+
+        if (user == null) {
```
Elena-Bruyako, line 21: use ternary operator instead of several if

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user: users) {
+            if (email.equals(user.getEmail())) {
+                return user;
+            }
+        }
+
```
Sanderfuck, line 23:
```suggestion
```

Sanderfuck: you can clear this line

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,16 @@ public class AuthenticationService {
...
+
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+
+        if (user == null) {
+            return false;
+        }
```
Sanderfuck, line 23: try to add this check like an additional condition after return, and it will look nice have only one return in this method

Sanderfuck (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
LiudmylaKuzmenko: I don't know how to add a check after the return but I made it in one return

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,19 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+
+        boolean canLogin = false;
+
+        if (user != null) {
+            canLogin = password.equals(user.getPassword());
+        }
+
+        return canLogin;
+
```
JJJazl, line 28:
```suggestion
return user != null && user.getPassword().equals(password);
```

JJJazl (changes requested):

JJJazl (approved):

PR #1138, opened 2024-11-23
 Implement a login method in AuthenticationService and findByEmail me…

RomanDidkivskyi:
…thod in UserService

### DIFF
--- a/src/main/java/mate/academy/service/UserService.java
+++ b/src/main/java/mate/academy/service/UserService.java
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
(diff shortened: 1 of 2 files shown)


Elena-Bruyako (changes requested):
build failed, fix it first

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
Elena-Bruyako, line 17: remove static 
```suggestion
    public User findByEmail(String email) {
```

Elena-Bruyako (changes requested):

PR #1139, opened 2024-11-27
Fix login and findByEmail methods

vitaliipp:


volodymyr-soltys97 (changes requested):
Hi, this task isn't frontend course, you need to switch to another course on the platform

PR #1141, opened 2024-11-29
Hw 0 oop

RomanDidkivskyi:
remove static


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,7 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User users = new UserService().findByEmail(email);
```
okuzan, line 16: userService should be private final field of the class, not a local unnamed variable

okuzan (changes requested):

PR #1143, opened 2024-12-03
login method and findByEmail method added

ira-levch:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User tempUser: users) {
```
```suggestion
        for (User user : users) {
```
use space before and after semicolon

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
...
      */
     public boolean login(String email, String password) {
+        UserService serv = new UserService();
+        User user = serv.findByEmail(email);
+        if (user != null) {
+            if (user.getPassword().equals(password)) {
+                return true;
+            }
+        }
+        System.out.println("No user with this e-mail is found");
```
okuzan, line 23: don't forget to remove sout (system.out.println) from your solution before committing your HW

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService serv = new UserService();
```
okuzan, line 16: don't shorten your local variables' names, be verbose! 🙏  Remember, developers spend much more time reading the code of others than writing their own, time saved by not typing out a few extra letters is not worth it

okuzan, line 16:
```suggestion
        UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
...
+        User user = serv.findByEmail(email);
+        if (user != null) {
+            if (user.getPassword().equals(password)) {
+                return true;
+            }
+        }
+        System.out.println("No user with this e-mail is found");
         return false;
```
okuzan, line 24: try to rewrite this solution without using boolean literals (true/false) at all. Use logical expressions directly in return

okuzan (changes requested):
Good job, some comments 👇

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
```
```suggestion
public class AuthenticationService {
    private final UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService newService = new UserService();
```
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
...
+        User user = newService.findByEmail(email);
+        if (user != null) {
+            if (user.getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
```
```suggestion
        return user != null && user.getPassword().equals(password));
```

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
ira-levch: I recceive the message: "Modifier 'private' is not allowed here"

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        final UserService newService = new UserService();
+        User user = newService.findByEmail(email);
+        return user != null && user.getPassword().equals(password));
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/AuthenticationService.java
Elena-Bruyako: There is no any error messages
Screenshot 2024-12-10 at 17 51 17

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User temporalUser : users) {
+            if (temporalUser.getEmail().equals(email)) {
+                return temporalUser;
```
Elena-Bruyako, line 20:
```suggestion
        for (User user : users) {
            if (user.getEmail().equals(email)) {
                return user;
```

PR #1144, opened 2024-12-04

wquieess:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User users = new User(email, password);
```
Elena-Bruyako, line 16: You should find user by calling method from UserService

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
...
+        if (users.getPassword().equals(password) && email != null) {
```
Elena-Bruyako, line 17: you can use ternary operator here

PR #1145, opened 2024-12-07
implemented methods dedicated to user logging in

gekatonheyr:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Elena-Bruyako, line 16: let's make UserService class-level variable to avoid creating this object each time you call login method

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user != null) {
```
Elena-Bruyako, line 18: you can use ternary operator here

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
gekatonheyr: Ok. I thought that since Java take care of automatically destruction of object (because it doesn't have explicitly defined destructors) - it would be right to create this object inside of method, just to renew users datarecords each time when method is called. Second reason - what will happen if between method calls to the User will be added new authentication records...

src/main/java/mate/academy/service/AuthenticationService.java
gekatonheyr: i tried something like this
return (this.userService.findByEmail(email) != null) ? this.userService.findByEmail(email).getPassword().equals(password) : false;
but stylecheck scolded me because of to long code line (((

Elena-Bruyako (approved):

PR #1147, opened 2024-12-09
Adding methods

sokoligor29:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,14 @@
...
-     * Imagine that some user wants to login to your site.
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Elena-Bruyako, line 7: it's better to make `userService` class-level variable

Elena-Bruyako (changes requested):

liliia-ponomarenko (approved):
Awesome 😎

PR #1151, opened 2024-12-15
update

PIpatii:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,21 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+public class AuthenticationService extends UserService {
```
liliia-ponomarenko, line 3: Don’t use inheritance here. if you want to use some method of UserService you should create an instance 
`private final UserService service = new UserService();`

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,21 @@
...
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (checkEmailAndPassword(password, email)) {
+            return true;
+        }
```
liliia-ponomarenko, line 17: simplify this method. Use `findByEmail` and if `user != null` compare passwords.

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,33 @@
...
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User tempUser : users) {
+            if (!tempUser.getEmail().equals(email)) {
+                break;
+            }
+        }
```
liliia-ponomarenko, line 23: this method returns null in any case. Is it expected behavior?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,33 @@
...
     }
+
+    public boolean checkEmailAndPassword(String password, String email) {
+        for (User tempUser : users) {
+            if (tempUser.getPassword().equals(password) && tempUser.getEmail().equals(email)) {
+                return true;
+            }
+        }
+        return false;
+    }
```
liliia-ponomarenko, line 34: remove this method

liliia-ponomarenko (changes requested):

PIpatii:
i fixed the mistakes

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,19 @@ public class AuthenticationService {
...
      * Return false in any other cases.
      */
+    private final UserService service = new UserService();
+
     public boolean login(String email, String password) {
+        if (service.findByEmail(email) != null) {
+            return service.checkingPassword(password) != null;
+        }
         return false;
+
```
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,33 @@
...
+        }
+        return null;
+    }
+
+    public User checkingPassword(String password) {
+        for (User tempUser : users) {
+            if (tempUser.getPassword().equals(password)) {
+                return tempUser;
+            }
+        }
```
liliia-ponomarenko, line 32: Duplicated code. remove this method and use `findByEmail` instead of this method

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,33 @@
...
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
             new User("bob@i.ua", "1234"),
             new User("alice@i.ua", "1234")
     };
 
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
```
liliia-ponomarenko, line 13:
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,33 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
```
liliia-ponomarenko, line 6:
```suggestion
    private static final User[] users = new User[] {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,19 @@ public class AuthenticationService {
...
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService service = new UserService();
+
     public boolean login(String email, String password) {
+        if (service.findByEmail(email) != null) {
+            return service.checkingPassword(password) != null;
```
liliia-ponomarenko, line 18: here you should check that the input password equals the password of the user that returns the  findByEmail method call

liliia-ponomarenko (changes requested):
Good job! Let's fix a few mistakes ;)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,18 @@ public class AuthenticationService {
...
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService service = new UserService();
+
     public boolean login(String email, String password) {
+        if (service.findByEmail(email) != null) {
+            return service.findByEmail(email).getPassword().equals(password);
+        }
         return false;
```
okuzan, line 20: you call findByEmal twice. save the result of the single call to a local variable, and reuse it. Try to use only one return in this method

okuzan (changes requested):

sarakhmen (approved):

PR #1153, opened 2024-12-19
add verify

FroGitHub:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
-        return null;
+    public static User findByEmail(String email) {
```
okuzan, line 17:
```suggestion
    public User findByEmail(String email) {
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
...
+        for (User u : users) {
```
```suggestion
        for (User user : users) {
```
avoid short one-letter naming, readability is important

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
...
+            if (email.equals(u.getEmail())) {
+                return u;
+            }
+        }
+        return new User("NoNe", "");
```
okuzan, line 23: what should you return, if there is no match? read the comment again carefully

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        return UserService.findByEmail(email)
+                .getPassword()
+                .equals(password);
```
okuzan, line 16: you will get NPE after refactoring UserService, so you'll need to make changes here too

okuzan (changes requested):

PR #1155, opened 2024-12-20
add verify

FroGitHub:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
...
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = this.userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        return user.getPassword().equals(password);
```
JJJazl, line 22: Just a hint) We can combine these checks together, but your solution can be more flexible for logging/debugging
```suggestion
        return user != null && user.getPassword().equals(password);
```

JJJazl (approved):

PR #1156, opened 2024-12-21
Solution

xMaXD:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -10,11 +10,19 @@ public class UserService {
...
+        for (User user : users) {
+            if (user.getEmail().equals(email)) {
+                return user;
+            }
+        }
         return null;
     }
 }
+
+
```
liliia-ponomarenko, line 28:
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,25 @@
...
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = new UserService().findByEmail(email);
```
liliia-ponomarenko, line 17: You should make the UserService class-level field 

liliia-ponomarenko (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
xMaXD: I changed and pushed new code, made UserService private final.

sarakhmen (approved):
lgtm!

PR #1158, opened 2024-12-23
add findByLogin and login foo

MykolaTurak:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
```
liliia-ponomarenko, line 18: Can you use a for-each loop?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
...
+            if (users[i].getEmail().equals(email)) {
+                return users[i];
+            }
+        }
         return null;
     }
+
```
liliia-ponomarenko, line 25:
```suggestion
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
liliia-ponomarenko, line 16: make it a class-level variable (private final)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
...
+        if (userService.findByEmail(email) == null) {
```
liliia-ponomarenko, line 17: Can you call the method findByEmail only once? Let’s  think about how to refactor this method

liliia-ponomarenko (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,16 @@ public class AuthenticationService {
...
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        if (user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
JJJazl, line 25: Let's simplify it) You can choose beetwen 2 versions
1)
```suggestion
        return user.getPassword().equals(password);
```
2) 
```suggestion
        return user != null && user.getPassword().equals(password);
```

JJJazl (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
...
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        return user != null && user.getPassword().equals(password);
```
JJJazl, line 22: You check that user != null above, so we don't need to do it again
```suggestion
        return user.getPassword().equals(password);
```

JJJazl (approved):

PR #1159, opened 2024-12-30
methods login and findByEmail implemented

ira-levch:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService newService = new UserService();
```
liliia-ponomarenko, line 16: better make it a class-level variable ;)

liliia-ponomarenko (changes requested):
Good job! One comment ;)

src/main/java/mate/academy/service/AuthenticationService.java
ira-levch: But in my code it is already made as a class-level variable:
        UserService newService = new UserService();
        User user = newService.findByEmail(email);
    }
}
Can you explain your comment in more details, please?

ira-levch: I've understood th issue, please, disregard

Elena-Bruyako: it means, that you shouldn't create `UserService newService = new UserService();` in `public boolean login(String email, String password)`, but create it as class-level variable

```

    private final UserService newService = new UserService();
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1160, opened 2025-01-03
implemented a login method

klv-liliya:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Elena-Bruyako, line 14: let's create UserService as class-level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
...
+        if (userService.findByEmail(email).getPassword() == password) {
```
Elena-Bruyako, line 15: you can use ternary operator here

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        return (userService.findByEmail(email) == null) ? false :
+                (userService.findByEmail(email).getPassword() == password) ? true : false;
```
sarakhmen, line 17: Here is another possible approach which I prefer more :)
```suggestion
        return userService.findByEmail(email) != null
                && userService.findByEmail(email).getPassword().equals(password);
```

sarakhmen (approved):
good job, you can improve your solution if you want :)

PR #1161, opened 2025-01-03
adding def's logic

l1kcs:
!!!Important!!!
 changed static final User[] users from **private** to **protected**

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,7 +3,7 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    protected static final User[] users = new User[] {
```
Elena-Bruyako, line 6:
```suggestion
    private static final User[] users = new User[] {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (User user : UserService.users) {
```
Elena-Bruyako, line 16: create UserService instance as class-level variable, find user by email and check if user by email exists and passed password is equal to user's password.

Elena-Bruyako (changes requested):

PR #1163, opened 2025-01-06
all tasks are solved

vooloshkaaa:
I implemented a login method in AuthenticationService and findByEmail method in UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
@@ -11,6 +13,17 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService newUserSerice = new UserService();
+        User user1 = newUserSerice.findByEmail(email);
+        if (user1 == null) {
+            return false;
+        } else {
+            String correctPassword = user1.getPassword();
+            if (password.equals(correctPassword)) {
+                return true;
+            } else {
+                return false;
+            }
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,17 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService newUserSerice = new UserService();
```
liliia-ponomarenko, line 16: better make it class level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,17 @@ public class AuthenticationService {
...
+        User user1 = newUserSerice.findByEmail(email);
+        if (user1 == null) {
+            return false;
+        } else {
+            String correctPassword = user1.getPassword();
+            if (password.equals(correctPassword)) {
+                return true;
+            } else {
+                return false;
+            }
```
liliia-ponomarenko, line 26: Avoid nesting conditional operators. One `if` block is enough, think about how to combine conditions and simplify your solution.

liliia-ponomarenko (changes requested):
good job! Let’s improve your solution ;)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService newUserSerice = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user1 = newUserSerice.findByEmail(email);
+        return user1 != null && password.equals(user1.getPassword());
```
sarakhmen, line 19:
```suggestion
        User user = newUserSerice.findByEmail(email);
        return user != null && password.equals(user.getPassword());
```
actually it's a bad practice to add such suffixes to the var names

sarakhmen (approved):
good job, you can improve your solution if you want ;)

PR #1166, opened 2025-01-14

stanko-dev:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,27 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private UserService userService = new UserService();
```
liliia-ponomarenko, line 6:
```suggestion
    private final UserService userService = new UserService();
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,28 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
```
liliia-ponomarenko, line 6:
```suggestion
    private static final User[] users = new User[] {
```

liliia-ponomarenko (approved):
Great! 😊

PR #1168, opened 2025-01-19
done

mixx1k:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,9 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
```
liliia-ponomarenko, line 5:
```suggestion
public class AuthenticationService {
```
Don’t use inheritance here. Create class-level variable 
` private final UserService userService = new UserService();`

liliia-ponomarenko (changes requested):
Good job! But let’s  fix one important thing ;)

liliia-ponomarenko (approved):
Awesome 😎

PR #1170, opened 2025-02-05
find user by email and check if they have access

vbedryk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako, line 16: it should be class-level variable, don't create `new UserService()` each time when `login()` is called

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
...
+        User foundUser = userService.findByEmail(email);
+        if (userService.findByEmail(email) != null && foundUser.getPassword().equals(password)) {
```
Elena-Bruyako, line 18: you can use return here

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user:users) {
```
Elena-Bruyako, line 18:
```suggestion
        for (User user : users) {
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1173, opened 2025-02-10
final

Woterix:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
```
Elena-Bruyako, line 11: remove

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (User user: UserService.getUsers()) {
```
Elena-Bruyako, line 16: create UserService as class-level variable
find user by email
return true if user by email exists and passed password is equal to user's password

Elena-Bruyako (changes requested):

PR #1174, opened 2025-02-10
Solved the task, passed all tests

IvanShevchenko1:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService service = new UserService();
```
Elena-Bruyako, line 16: it should be class-level variable, don't need create new UserService() each time you call login method

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        if (service.findByEmail(email) != null) {
+            User foundUser = service.findByEmail(email);
+            return password.equals(foundUser.getPassword());
+        }
```
Elena-Bruyako, line 20:
```suggestion
        User user = userService.findByEmail(email);
        return user != null && user.getPassword().equals(password);
```

Elena-Bruyako (changes requested):

PR #1175, opened 2025-02-11
change method login in clas AuthenticationService and change method f…

b1gsal:
…indByEmail in clas UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -10,7 +10,17 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        if (userService.findByEmail(email) == null) {
+            return false;
+        }
+        if (userService.findByEmail(email).getEmail().equals(email)) {
+            if (userService.findByEmail(email).getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,17 @@ public class AuthenticationService {
...
+
     public boolean login(String email, String password) {
+        if (userService.findByEmail(email) == null) {
+            return false;
+        }
+        if (userService.findByEmail(email).getEmail().equals(email)) {
+            if (userService.findByEmail(email).getPassword().equals(password)) {
+                return true;
+            }
+        }
```
sarakhmen, line 23: findByEmail() method is called several times. Can we call it only once and assign the result to the variable so that you can reuse it?

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
+            if (email.equals(users[i].getEmail())) {
+                return users[i];
+            }
+        }
```
sarakhmen, line 22: let's use foreach loop instead

sarakhmen (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,18 @@ public class AuthenticationService {
...
+        User byEmail = userService.findByEmail(email);
+        if (byEmail == null) {
+            return false;
+        }
+        if (byEmail.getEmail().equals(email)) {
+            if (byEmail.getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
```
```suggestion
        User user = userService.findByEmail(email);
        return user != null && user.getPassword().equals(password);
```

Elena-Bruyako (approved):

PR #1178, opened 2025-02-17
implemented login() and findByEmail() methods

DZotsenko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private static final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        return userService.findByEmail(email) != null && userService.findByEmail(email)
```
liliia-ponomarenko, line 16: Can you call the method findByEmail only once? Don’t complicate the performance of your solution ;)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private static final UserService userService = new UserService();
```
liliia-ponomarenko, line 13:
```suggestion
    private final UserService userService = new UserService();
```

liliia-ponomarenko (changes requested):

JJJazl (approved):
Great! Let's move on)

PR #1181, opened 2025-02-25
Hw1 login auth

LoranV:
Finished task. Modified login fun. Modified findByEmail func.

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,16 +1,13 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
-    /**
-     * Imagine that some user wants to login to your site.
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+        if(user != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,12 @@ public class UserService {
...
     };
 
-    /**
-     * Find user by email. All users are stored in <code>private static final User[] users</code>
-     * @param email - the input parameter
-     * @return - user if his email is equal to passed email.
-     * Return <code>null</code> if there is no suitable user
-     */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
liliia-ponomarenko, line 11: Don’t  make the method static

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,13 @@
...
-     * Imagine that some user wants to login to your site.
-     * You should check if user credentials (login and password) are valid or not.
-     * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
-     * @param password - user's password
-     * @return true if user by email exists and passed password is equal to user's password.
-     * Return false in any other cases.
-     */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
```
liliia-ponomarenko, line 7: Make UserService private final class level variable

liliia-ponomarenko (changes requested):
Good job! A few comments ;)

src/main/java/mate/academy/service/AuthenticationService.java
LoranV: Hello, sorry, Isn't it better without creating new variable, if we can use class without declaring?

PR #1182, opened 2025-02-25
Login success

ynazarets:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
liliia-ponomarenko, line 16: Make UserService private final class-level variable ;)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        return user.getPassword().equals(password);
```
liliia-ponomarenko, line 18: Can the user be null? What happens if the user null value? Let’s handle it ;)

liliia-ponomarenko (changes requested):

JJJazl (approved):
Great! Let's move on)

PR #1183, opened 2025-02-26
update methods login and findByEmail

JackWillov:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
liliia-ponomarenko, line 14: make it a private final class-level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,13 @@ public class AuthenticationService {
...
+        if (userService.findByEmail(email) == null) {
```
liliia-ponomarenko, line 15: Can you call the method findByEmail only once?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,13 @@ public class AuthenticationService {
...
+            return false;
+        } else if (userService.findByEmail(email).getEmail().equals(email)
```
liliia-ponomarenko, line 17: do you really need to check the email parameter?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,13 @@ public class AuthenticationService {
...
+                && userService.findByEmail(email).getPassword().equals(password)) {
+            return true;
```
liliia-ponomarenko, line 19: Think about how to refactor it and avoid if-else construction

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (int i = 0; i < users.length; i++) {
```
liliia-ponomarenko, line 18: can you use for-each loop?

liliia-ponomarenko (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
JackWillov: of course. But I have 1 question, do I need to respond to the comments here if I agree and update this part of code. For example (yes, I did, of course). This is my first assignment which review by a mentor.

liliia-ponomarenko: no, you don't have to, just fix the comments. If something seems unclear, it is better to write a question in the general chat on the platform to get a quick answer

liliia-ponomarenko (approved):
Awesome 😎

PR #1184, opened 2025-02-27
Done

Hot00Tea:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako, line 16: Let's make it class-level variable to avoid creating new UserService each time when login method is called

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
...
+
+        if (email == null || password == null) {
```
Elena-Bruyako, line 18: remove this checking

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
...
+            return false;
+        }
+
+        User user = userService.findByEmail(email);
+
```
Elena-Bruyako, line 23: remove redundant empty line
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,15 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        if (email == null) {
```
Elena-Bruyako, line 18: remove

Elena-Bruyako (changes requested):

PR #1187, opened 2025-03-24
do progr

romander333:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -1,20 +1,25 @@
...
-    };
+    private List<User> users;
+
+    public UserService() {
+        users = List.of(
+                new User("bob@i.ua", "1234"),
+                new User("alice@i.ua", "1234"),
+                new User("john@i.ua", "5678")
+        );
+    }
```
ahoienko, line 15: What is the reason to remove pre-created users array and adding list with additional user?

ahoienko (changes requested):
Nice, left one question

Elena-Bruyako (approved):

PR #1190, opened 2025-03-31
implemented findByEmail method in UserService and login method in Aut…

yaaariiik:
…henticationService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,6 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private static final UserService USER_SERVICE = new UserService();
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,6 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
+    private static final UserService USER_SERVICE = new UserService();
```
JJJazl, line 6:
```suggestion
    private final UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +13,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+        User user = USER_SERVICE.findByEmail(email);
+        if (user != null) {
+            return user.getPassword().equals(password);
+        }
         return false;
```
JJJazl, line 22: This approach is better for debugging/logging purposes, but just so you know you can do it this way too
`return user != null && user.getPassword().equals(password);`

JJJazl (changes requested):
Good! Let's improve your solution)

src/main/java/mate/academy/service/AuthenticationService.java
yaaariiik: Thanks for suggestion! It looks realy better.

Elena-Bruyako (approved):

PR #1193, opened 2025-04-16
added new logic for functions "login" and "findByEmail"

BohachenkoArtem:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+
+        UserService service = new UserService();
```
Elena-Bruyako, line 15: it's better to make this variable class-level

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,14 @@ public class AuthenticationService {
...
+
+        if (service.findByEmail(email) == null) {
```
Elena-Bruyako, line 17: try to avoid calling `findByEmail` method twice

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,17 @@ public class AuthenticationService {
...
-        return false;
+
+        User someUser = service.findByEmail(email);
+
+        if (someUser == null) {
+            return false;
+        } else {
+            return someUser.getPassword().equals(password);
+        }
+
```
JJJazl, line 26:
```suggestion
        User user = service.findByEmail(email);
        
        // First approach (better for logging/debug purposes)
        if (user == null) {
            return false;
        }
        return user.getPassword().equals(password);
        
        // Second approach (less lines of codes, easier to read)
        return user != null && user.getPassword().equals(password);
        }
```

JJJazl (changes requested):
Good! Let's improve your solution)

JJJazl (approved):
Please, pay attention to the variable names

PR #1196, opened 2025-04-26
feature_1

alex2850:


Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako: it should be class-level variable, don't create new object each time when method login is called

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1213, opened 2025-05-18
add findByEmail and login method

DaniilShelofast:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,19 @@ public class AuthenticationService {
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+
+        UserService userService = new UserService();
```
liliia-ponomarenko, line 16: better make private final class-level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -5,12 +5,19 @@ public class AuthenticationService {
...
+        if (userService.findByEmail(email).getEmail().equals(
+                userService.findByEmail(password).getPassword())) {
```
liliia-ponomarenko, line 18: Can you call the method findByEmail only once? Can this code provide NPE?

liliia-ponomarenko (changes requested):
Let’s improve your solution ;)

DaniilShelofast:
I know this solution is not correct, I did it but push git failed. I made a
getUsers for the users array, so I could get the users for the login method


пн, 19 трав. 2025 р. о 10:25 Liliia Ponomarenko ***@***.***>
пише:

PR #1221, opened 2025-06-04
feat: implement login and findByEmail methods

RedFooEd:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = new UserService().findByEmail(email);
```
liliia-ponomarenko, line 16: Make UserService class level variable

liliia-ponomarenko (approved):
Good job! One comment;)

PR #1229, opened 2025-06-17
Implement login method in AuthenticationService and findByEmail in Us…

MaritaPasha:
…erService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,16 +1,23 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
      * All users are stored in <code>UserService</code> class.
+     *
      * @param email - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    private static final User[] users = new User[]{
```
sarakhmen: this cnahge should be reverted

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,24 @@
...
             new User("bob@i.ua", "1234"),
             new User("alice@i.ua", "1234")
     };
 
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
```
sarakhmen: and this as well

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,23 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
      * All users are stored in <code>UserService</code> class.
+     *
```
sarakhmen: and here

sarakhmen (changes requested):
good job! just a minor fix needed

src/main/java/mate/academy/service/UserService.java
MaritaPasha: Thanks! I fixed it. 78e24262eb5601bb69e555fccd054e0bcf6a30e0

PR #1230, opened 2025-06-18
Implement login and findByEmail methods

ArtemLysyuk:


Elena-Bruyako (changes requested):
let's fix build first

ArtemLysyuk:
@Elena-Bruyako fixed

PR #1231, opened 2025-06-19
fin

OleksiiKolinko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private static final UserService userService = new UserService();
```
```suggestion
    private final UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,13 @@ public class AuthenticationService {
...
+
     public boolean login(String email, String password) {
-        return false;
+        final User user = userService.findByEmail(email);
+        if (user == null) {
```
Elena-Bruyako: try to combine into one return without if

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +16,8 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        return Arrays.stream(users)
```
Elena-Bruyako: don't use stream

Elena-Bruyako (changes requested):

liliia-ponomarenko (approved):
LGTM!

PR #1233, opened 2025-07-06
implement methods for users verify

DoctorJambo:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public static User[] getUsers() {
+        return users;
+    }
+
```
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
...
     /**
      * Find user by email. All users are stored in <code>private static final User[] users</code>
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public User findByEmail(String email, User[] users) {
```
```suggestion
    public User findByEmail(String email {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (User user : UserService.getUsers()) {
```
Elena-Bruyako: Create class-level variable of `UserService`

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+            if (email.equals(user.getEmail()) && password.equals(user.getPassword())) {
```
Elena-Bruyako: use ternary operator here and return directly

Elena-Bruyako (changes requested):
looks like your have project inside project, don't change the packaging, just implement task

liliia-ponomarenko (approved):
Good job!

PR #1238, opened 2025-07-24
Added solution for AuthenticationService.java and UserService.java

Januzao:
Added solution for AuthenticationService.java and UserService.java

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -12,17 +12,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        UserService userService = new UserService();
         User user = userService.findByEmail(email);
-        if (user == null) {
-            return false;
-        }
 
-        if (user.getPassword().equals(password)) {
-            return true;
-        }
-
-        return false;
+        return email != null && user != null && user.getPassword().equals(password);
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Elena-Bruyako: create UserService as a class-level variable to avoid creating instance each time you call login method

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,16 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user == null) {
```
Elena-Bruyako: use ternary operator instead of several if

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,16 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
-        return null;
+        User bob = users[0];
```
Elena-Bruyako: use for loop to find use by email in `users[]`

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,12 @@ public class AuthenticationService {
...
      * Return false in any other cases.
      */
+
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+
+        return email != null && user != null && user.getPassword().equals(password);
```
sarakhmen: email check is redundant here, please remove it

sarakhmen (changes requested):
good job, but let's slightly improve your solution :)

liliia-ponomarenko (approved):
Awesome 😎

PR #1240, opened 2025-07-29
hw jv-oop-lesson-0 solution

Vitalii85:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
liliia-ponomarenko:
```suggestion
    publiс User findByEmail(String email) {
```
Don't make this method static

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,7 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        return (UserService.findByEmail(email) == null) ? false
+                : UserService.findByEmail(email).getPassword().equals(password);
```
liliia-ponomarenko: Make UserService class-level variable. Call the method findByEmail only once

liliia-ponomarenko (changes requested):
Good job! A few comments ;)

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako: it should be class-level variable

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,10 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private static final UserService userService = new UserService();
```
Elena-Bruyako: never create static mutable object
```suggestion
    private final UserService userService = new UserService();
```

Elena-Bruyako (changes requested):
one comment

Elena-Bruyako (approved):

PR #1242, opened 2025-08-03
Adding system of finding and access email and password

S0rbex:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,7 +1,9 @@
 package mate.academy.service;
 
-public class AuthenticationService {
-    /**
+import mate.academy.model.User;
+
+public class AuthenticationService extends UserService {
```
```suggestion
public class AuthenticationService {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = findByEmail(email);
```
Elena-Bruyako: create `private final UserService userService = new UserService();` class-level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
...
+        if (user == null) {
+            return false;
+        } else {
+            if (user.getPassword().equals(password)) {
+                return true;
+            }
+        }
         return false;
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
S0rbex: Yeah, already should be done, thanks!

Elena-Bruyako (approved):

PR #1244, opened 2025-08-07
implemented login and findByEmail

Inna1989F:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
...
-     * @param email - user's email
+     *
+     * @param email    - user's emai
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako, line 10: ```suggestion```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
```
Elena-Bruyako: make UserService class-level variable to avoid creating object each time you call `login` method
```suggestion
    private final UserService userService = new UserService();
    /**
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
...
+     * @param email    - user's emai
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user == null) {
```
Elena-Bruyako: you can use ternary operator here

Elena-Bruyako (changes requested):

Inna1989F (commented):
Fixed

PR #1245, opened 2025-08-07
add implementations of login and findByEmail methods

FranticSkipper:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
```
```suggestion
public class AuthenticationService {
    private final UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako: it's better to make it class-level variable to avoid creating new `UserService` each time you call `login` method
```suggestion
```

Elena-Bruyako (changes requested):

FranticSkipper:
i have added change in 8bdfc78513fec18bd857e59e27e4eb21e7c0456d

Elena-Bruyako (approved):

PR #1247, opened 2025-08-18
solved task with creating implementation of needed methods

tsuki99:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = new UserService().findByEmail(email);
```
Elena-Bruyako: make `UserService` as a class-level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
...
+
+        return new UserService().findByEmail(email) != null
```
```suggestion
        return userService.findByEmail(email) != null
```

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
```
```suggestion
public class AuthenticationService {
    private final UserService userService = new UserService();
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
```suggestion
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):
Good job!

PR #1251, opened 2025-09-02
Created login() and findByEmail() methods implementation

noleynik29:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Elena-Bruyako: make `UserService userService` class-level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user != null) {
```
Elena-Bruyako: use ternary operator here

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1255, opened 2025-09-12
added methods to implement authentication

yurakukh:
Added simple authentication methods


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
sarakhmen: move this service and its initialization to the class fields

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
...
+        if (userService.findByEmail(email) != null) {
+            return userService.findByEmail(email).getPassword().equals(password);
```
sarakhmen: findByEmail method is called twice, let's fix it

sarakhmen (changes requested):
let's improve your solution :)

yurakukh:
Solved problems that mentor told me to

Elena-Bruyako (approved):

PR #1257, opened 2025-09-22
Solution

Woteraz:
problems solved


Woteraz:
Fixed "Login " method. Was wrong conditions. Did mvn clean package, build success.

PR #1258, opened 2025-09-23
сomplete authentication service task

Freemankon:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako: make UserService class-level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User foundUser = userService.findByEmail(email);
+        if (foundUser == null) {
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1259, opened 2025-09-23
Implemented a login method

NaKraVon:
Implemented a login method in AuthenticationService and findByEmail method in UserService


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,13 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
```suggestion
    public User findByEmail(String email) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+
+        if (UserService.findByEmail(email) != null) {
```
Elena-Bruyako: create UserService as class-level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,10 @@ public class AuthenticationService {
...
+            return UserService.findByEmail(email).getPassword().equals(password);
```
Elena-Bruyako: don't call findByEmail twice, create local variable for result of the method

Elena-Bruyako (changes requested):

src/main/java/mate/academy/service/UserService.java
NaKraVon: Fixed in 896e8b2f230cc2a792f7cb997551e625cc4bb7a3

src/main/java/mate/academy/service/AuthenticationService.java
NaKraVon: Created a local variable "user" in 896e8b2f230cc2a792f7cb997551e625cc4bb7a3

Elena-Bruyako (approved):

PR #1265, opened 2025-10-15
added solution for methods

SvitlanaLychagina:
added solution for login method  in AuthenticationService and findByEmail method in UserService


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Elena-Bruyako: make it class-level variable

Elena-Bruyako (changes requested):
Good job! One comment

src/main/java/mate/academy/service/AuthenticationService.java
SvitlanaLychagina: thanks, fix this at 8efc0c7615f1f621014ef807e430a46aa8f56263

sarakhmen (approved):

PR #1271, opened 2025-11-20
have done all required tasks

Polya555:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,18 @@ public class AuthenticationService {
...
+        this.userService = new UserService();
+    }
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        return password != null && password.equals(user.getPassword());
```
sarakhmen: use ternary operator instead. Also you can make something like user.getPassword().equals(password) and skip password != null check this way

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
...
     public User findByEmail(String email) {
+        for (User user : users) {
+            if (email != null && email.equals(user.getEmail())) {
+                return user;
+            }
+        }
         return null;
     }
 }
+
```
sarakhmen:
```suggestion
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user : users) {
+            if (email != null && email.equals(user.getEmail())) {
```
sarakhmen: same here, we can skip email != null check if use equals the opposite way

sarakhmen (changes requested):

src/main/java/mate/academy/service/AuthenticationService.java
Polya555: 5eaf08e934b90cb54459437f4f48b9bf1ed21190

Polya555: ebc9b06be3886948b9b246456020098af444094b

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,18 @@ public class AuthenticationService {
...
+    public AuthenticationService() {
+        this.userService = new UserService();
+    }
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null) {
+            return user.getPassword().equals(password) ? true : false;
+        }
         return false;
```
```suggestion
       return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,18 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
+    private UserService userService;
+
+    public AuthenticationService() {
+        this.userService = new UserService();
+    }
```
Elena-Bruyako, line 20:
```suggestion
    private final UserService userService;
```

Elena-Bruyako (changes requested):
see comments

src/main/java/mate/academy/service/AuthenticationService.java
Polya555: this change makes this error:  variable userService not initialized in the default constructor

sarakhmen (approved):
lgtm!

PR #1272, opened 2025-12-02
Solutions

kastashenkova:


kastashenkova:
added solutions

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,7 +3,7 @@
 import mate.academy.model.User;
 
 public class UserService {
-    private static final User[] users = new User[] {
+    static final User[] users = new User[] {
```
```suggestion
    private static final User[] users = new User[] {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        for (mate.academy.model.User user : UserService.users) {
+            if (email.equals(user.getEmail()) && password.equals(user.getPassword())) {
+                return true;
+            }
+        }
         return false;
```
```suggestion
        User user = userService.findByEmail(email);
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
```
Elena-Bruyako, line 17: create UserService as class-level variable

Elena-Bruyako (changes requested):
see comments

kastashenkova:
Please, check if I corrected it right!

Elena-Bruyako (approved):

PR #1273, opened 2025-12-06
added solution for class UserService and AuthenticationService

zayatstarasuk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,20 @@
...
+    public AuthenticationService() {
+        this.userService = new UserService();
+    }
+
     public boolean login(String email, String password) {
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
+            return true;
+        }
         return false;
```
sarakhmen, line 18: use ternary operator instead

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,12 @@ public class UserService {
...
     };
 
-    /**
-     * Find user by email. All users are stored in <code>private static final User[] users</code>
-     * @param email - the input parameter
-     * @return - user if his email is equal to passed email.
-     * Return <code>null</code> if there is no suitable user
-     */
     public User findByEmail(String email) {
+        for (User user:users) {
```
sarakhmen, line 12:
```suggestion
        for (User user : users) {
```

sarakhmen (changes requested):
good job! But let's slightly improve your solution :)

PR #1275, opened 2025-12-17
added login and findByEmail methods

Viodon:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,14 @@ public class AuthenticationService {
...
      */
+    private UserService user = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        if (user.findByEmail(email) == null) {
+            return false;
+        }
+        return email.equals(user.findByEmail(email).getEmail())
+            && password.equals(user.findByEmail(email).getPassword());
```
sarakhmen: findByEmail method is called 3 times. Let's call it once and store the result in a local variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,14 @@ public class AuthenticationService {
...
      * Return false in any other cases.
      */
+    private UserService user = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        if (user.findByEmail(email) == null) {
+            return false;
+        }
+        return email.equals(user.findByEmail(email).getEmail())
```
sarakhmen: no need to validate that found user email is equal to the provided one

sarakhmen (changes requested):
let's slightly improve your solution :)

Elena-Bruyako (approved):

PR #1281, opened 2026-01-06
solution of exercise

kaminskyivladyslav:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,32 @@
...
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user == null || email == null || password == null) {
+            return false;
+        } else if (user.getPassword().equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -8,13 +8,22 @@ public class UserService {
             new User("alice@i.ua", "1234")
     };
 
+    public void setUser(User user) {
+
+    }
+
```
```suggestion
```

Elena-Bruyako (changes requested):
see comments

src/main/java/mate/academy/service/UserService.java
kaminskyivladyslav: was fixed

src/main/java/mate/academy/service/AuthenticationService.java
kaminskyivladyslav: fixed

kaminskyivladyslav:
fixed changes

Elena-Bruyako (approved):

PR #1282, opened 2026-01-26
implemented methods

AndriiAlieksieienko:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,17 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+
+        boolean isPermitted = false;
```
don't need to use two if operators and remove redundant boolean isPermitted variable

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,14 @@ public class UserService {
...
-        return null;
+        User searchedUser = null;
+
+        for (User user : users) {
+            if (user.getEmail().equals(email)) {
+                searchedUser = user;
+            }
+        }
+
+        return searchedUser;
```
```suggestion
        for (User user : users) {
            if (user.getEmail().equals(email)) {
                return user;
            }
        }
        return null;
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,17 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
```suggestion
        UserService userService = new UserService();
        User user = userService.findByEmail(email);
        return user != null && user.getPassword().equals(password);

**Review by Elena-Bruyako — CHANGES_REQUESTED** [review]
see comments

File: src/main/java/mate/academy/service/AuthenticationService.java [hunk 4]
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+
+        return (user != null) && (user.getPassword().equals(password));
```
fetis-off, line 19: minor: 
```suggestion
        return user != null && user.getPassword().equals(password);
```

fetis-off (changes requested):
Good one! But It seems to me you forget to resolve the second comment. Please, do it

Elena-Bruyako (approved):

PR #1283, opened 2026-01-27
solve task in AuthenticationService.java and UserService.java

MrRaveGit:


fetis-off (approved):
Good one, approved :)

PR #1284, opened 2026-02-01
modification user authentications

wiwalentusr-ux:
modification user authentications. modificate class AuthenticationService  & UserService

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,5 +1,7 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
```
fetis-off, line 20: if user's email is null -> ?
should be validated

the same for password

fetis-off (changes requested):
Good job! See coment

src/main/java/mate/academy/service/AuthenticationService.java
wiwalentusr-ux: Thanks for the note, null check added a18983d322d17edd5db1da5d2b349c42f62de035

fetis-off: Just small tip for the future: we have method in` org.apache.commons.lang3 ` package `StringUtils.isEmpty(String s)`
I use it every time for string validation :)

fetis-off (approved):
Good one!

PR #1286, opened 2026-02-02
Authentication lesson 0 

SofiiaAsieieva:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
...
      */
     public User findByEmail(String email) {
-        return null;
+        User result = null;
+        for (User user: users) {
+            if (user.getEmail().equals(email)) {
+                result = user;
+            }
+        }
+        return result;
```
```suggestion
        for (User user: users) {
            if (user.getEmail().equals(email)) {
                return user;
            }
        }
        return null;
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
```
fetis-off, line 21: What if user email is `null`? Better to add this validation on email and password.

fetis-off (changes requested):
Good! Left a few comments so let's improve your solution

PR #1287, opened 2026-02-04
My solution

plvkIvan:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user : users) {
+            if (user.getEmail().equals(email)) {
```
fetis-off, line 23: What if User email is null?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,14 @@ public class AuthenticationService {
...
+
+    private UserService userService = new UserService();
+
     public boolean login(String email, String password) {
-        return false;
+        User user = userService.findByEmail(email);
+        if (user == null) {
+            return false;
+        }
+        return user.getPassword().equals(password);
```
fetis-off, line 23: What if User password is null?

fetis-off (changes requested):
Please, add proper null handling validation

Elena-Bruyako (approved):

PR #1291, opened 2026-02-16
Hom solution

yarroslaw3842o84023984902:
https://github.com/yarroslaw3842o84023984902/jv-oop-lesson-0

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,4 +1,5 @@
 package mate.academy.service;
+
 import mate.academy.model.User;
 
 public class AuthenticationService {
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,16 +1,25 @@
 package mate.academy.service;
+import mate.academy.model.User;
 
 public class AuthenticationService {
     /**
      * Imagine that some user wants to login to your site.
      * You should check if user credentials (login and password) are valid or not.
      * All users are stored in <code>UserService</code> class.
-     * @param email - user's email
+     *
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,12 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User el : users) {
+            if (el.getEmail().equals(email)) {
```
fetis-off, line 22: what if user email is null?

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,26 @@
...
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+
+        if (user == null) {
+            return false;
+        }
+
+        return user.getPassword().equals(password);
```
fetis-off, line 28: what if user password is null?

fetis-off (changes requested):
Left a few comments. Please add null validation

src/main/java/mate/academy/service/UserService.java
yarroslaw3842o84023984902: Hi. Fixed 8904b6e849bdb8bd4b2bde5ed07fa3821b33107f

src/main/java/mate/academy/service/AuthenticationService.java
yarroslaw3842o84023984902: Hi, fiexed 8904b6e849bdb8bd4b2bde5ed07fa3821b33107f 
@fetis-off

fetis-off (approved):
Good one

PR #1292, opened 2026-02-20
solution

IvanGuculyak:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,13 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
+        for (User user: users) {
+            if (user.getEmail().equals(email)) {
```
fetis-off, line 23: what if user email is null? 
Let's add validation for this case

fetis-off (changes requested):
Good work! Left one comment

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,17 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
```suggestion
    public User findByEmail(String email) {
```
try to avoid using static methods

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +11,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        if (UserService.findByEmail(email) != null) {
+            return UserService.findByEmail(email).getPassword().equals(password);
+        }
         return false;
     }
 }
```
fetis-off, line 25:
```suggestion
public boolean login(String email, String password) {
UserService userService = new UserService();
User user = userService.findByEmail(email);
        if (user != null) {
            return user.getPassword().equals(password);
        }
return false;
}
}
```

But the question now: What is user password is null?

fetis-off, line 25:
```suggestion
public boolean login(String email, String password) {
UserService userService = new UserService();
User user = userService.findByEmail(email);
        if (user != null) {
            return user.getPassword().equals(password);
        }
return false;
}
}
```

But the question now: What if user password is null?

fetis-off (changes requested):
See comment and add validation for password field

fetis-off (approved):
Good

PR #1293, opened 2026-02-24
Feature/login

Krupkolllia:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,7 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        User user = new UserService().findByEmail(email);
```
Elena-Bruyako: create class-level variable for UserService

Elena-Bruyako (changes requested):
good job
fix one comment

Elena-Bruyako (approved):

PR #1295, opened 2026-02-28
implement simple auth

AlexanderShpilka:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,9 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako, line 16: fot the future, it's better to make UserService as class-level variable to avoid creating it each time you call login method

Elena-Bruyako (approved):
good job
see comments for the future

src/main/java/mate/academy/service/AuthenticationService.java
AlexanderShpilka: Got it. Thank you for the comment.

PR #1296, opened 2026-03-06
Fix logic in UserService and AuthenticationService

RodionPlakhotniuk:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+
+        User user = new UserService().findByEmail(email);
```
fetis-off: lets better create a separate userService instance and call it.
```
userService.findByEmail(email); 
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,28 @@
...
      * Find user by email. All users are stored in <code>private static final User[] users</code>
+     *
      * @param email - the input parameter
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
+
     public User findByEmail(String email) {
+
+        for (User userEmail : UserService.users) {
```
```suggestion
        for (User user : UserService.users) {
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,28 @@
...
+
+            if (email.equals(userEmail.getEmail())) {
+                return userEmail;
+            }
```
fetis-off: What if user email is null? Lets handle it
```suggestion
            if (email.equals(userEmail.getEmail())) {
                return user;
            }
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,14 @@ public class AuthenticationService {
...
+
+        if (user != null && password.equals(user.getPassword())) {
```
fetis-off: what if user password is null?

fetis-off (changes requested):
Good work! See comments

src/main/java/mate/academy/service/UserService.java
```diff
@@ -3,18 +3,32 @@
...
      */
+
     public User findByEmail(String email) {
+
+        if (email == null) {
+            return null;
+        }
+
+        for (User user : UserService.users) {
+
```
fetis-off, line 26:
```suggestion
```
redundant empty line

fetis-off (approved):
Approved

PR #1297, opened 2026-03-12
implement findByEmail and login methods

MitiD99:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +12,16 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
+
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user == null) {
```
Elena-Bruyako, line 20: use ternary operator

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,14 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+
+        for (User user : users) {
+            if (users.length == 0) {
```
Elena-Bruyako, line 20: remove if else

Elena-Bruyako (changes requested):
Good job
See comments

PR #1299, opened 2026-03-16
solution for Implement Simple Authentication

MartaRusanovska:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+public class AuthenticationService extends UserService {
```
```suggestion
public class AuthenticationService {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
...
-     * @param email - user's email
+     *
+     * @param email    - user's email
      * @param password - user's password
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        var user = findByEmail(email);
+        if (user == null) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -1,16 +1,24 @@
 package mate.academy.service;
 
-public class AuthenticationService {
+public class AuthenticationService extends UserService {
     /**
```
Elena-Bruyako, line 6: make UserService as class-level variable

Elena-Bruyako (changes requested):
Good job
See comments

src/main/java/mate/academy/service/AuthenticationService.java
MartaRusanovska: Pls check fixed PR .

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,11 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+        for (User user : users) {
+            if (email.equals(user.getEmail())) {
```
fetis-off, line 19: what if email parameter is null? Let's better add this check

fetis-off (changes requested):
Good one! Left one comment

fetis-off (approved):
approved

PR #1300, opened 2026-03-26
Ht jv oop lesson 0

n1ghtmare88:


src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
      * @return - user if his email is equal to passed email.
      * Return <code>null</code> if there is no suitable user
      */
-    public User findByEmail(String email) {
+    public static User findByEmail(String email) {
```
Elena-Bruyako: Why did you add static here?
```suggestion
    public User findByEmail(String email) {
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -14,7 +14,12 @@ public class UserService {
...
+        for (User user:users) {
```
```suggestion
        for (User user : users) {
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
+        if (user != null
```
Elena-Bruyako: use ternary operator instead of ` if `

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        User user = UserService.findByEmail(email);
```
Elena-Bruyako: UserService should be class-level variable

Elena-Bruyako (changes requested):
See comments

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +14,11 @@ public class AuthenticationService {
...
      * Return false in any other cases.
      */
+
     public boolean login(String email, String password) {
-        return false;
+        user = userService.findByEmail(email);
+        return (user != null
+                && email.equals(user.getEmail())
+                && password.equals(user.getPassword())) ? true : false;
     }
```
Elena-Bruyako, line 21:
```suggestion
    private UserService userService = new UserService();
/**
    * Imagine that some user wants to login to your site.
    * You should check if user credentials (login and password) are valid or not.
@@ -10,7 +14,11 @@ public class AuthenticationService {
    * @return true if user by email exists and passed password is equal to user's password.
    * Return false in any other cases.
    */

public boolean login(String email, String password) {
        User user = userService.findByEmail(email);
        return user != null && user.getPassword().equals(password);
}
```

Elena-Bruyako (changes requested):

Elena-Bruyako (approved):

PR #1302, opened 2026-04-10
implement a login method in AuthenticationService and findByEmail met…

Sasha-Vradiy:
…hod in UserService


Elena-Bruyako (approved):
Good job

PR #1304, opened 2026-04-26

yashchuk-t:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,12 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+
+        if (user != null && user.getPassword().equals(password)) {
```
fetis-off: what if password is null?

fetis-off (changes requested):
See comment

Elena-Bruyako (approved):

PR #1305, opened 2026-04-27
solution solved

Tacha-hub:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,13 @@ public class AuthenticationService {
...
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+
+        if (user == null) {
+            return false;
+        }
+
+        return user.getPassword().equals(password);
```
fetis-off, line 27: what if user password is null? Let's add this check

fetis-off (changes requested):
Good work! See comment

src/main/java/mate/academy/service/AuthenticationService.java
Tacha-hub: > what if user password is null? Let's add this check

done, should I make a new pull request? I fixed the code and added_commit&push_them

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,17 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
```
Elena-Bruyako, line 16: UserService should be class-level variable, to avoid creating object each time you call method

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,17 @@ public class AuthenticationService {
...
+
+        if (user == null) {
+            return false;
+        }
+
+        if (password == null) {
+            return false;
+        }
+
+        return user.getPassword().equals(password);
```
Elena-Bruyako, line 27: you can simplify code
```suggestion
        return password != null && user != null && user.getPassword().equals(password);
```

Elena-Bruyako (approved):
good job
see comments for the future

PR #1306, opened 2026-05-01
Implement methods findByEmail and login

Vladyslav19ua:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
-        return false;
+        UserService userService = new UserService();
+        User userServiceByEmail = userService.findByEmail(email);
```
fetis-off, line 21:
```suggestion
        User user = userService.findByEmail(email);
```

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,8 @@ public class AuthenticationService {
...
+        return userServiceByEmail != null && userServiceByEmail.getPassword().equals(password);
```
fetis-off: what if user password or email is null? Let's add this checks 
If at least one of these fields is null, then return false.

fetis-off (changes requested):
Good work. See comments

src/main/java/mate/academy/service/AuthenticationService.java
Vladyslav19ua: Hi! I made the edits suitable to the recommendation

fetis-off (approved):
good!

PR #1307, opened 2026-05-09
created two methods login and findByEmail

acidden:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
+        User user = userService.findByEmail(email);
+        if (user != null && user.getPassword().equals(password)) {
```
Elena-Bruyako, line 18: here you can use ternary operator

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Elena-Bruyako, line 16: to avoid creating `new UserService `each time you call `login` method, it's better to make it class-level variable

Elena-Bruyako (approved):
good job
see comments for the future

src/main/java/mate/academy/service/AuthenticationService.java
acidden: return user != null && user.getPassword().equals(password);

PR #1308, opened 2026-05-18
Implement simple authentication service with login validation

Vladyslav881:
Changes
- Implemented login method in AuthenticationService
- Implemented findByEmail method in UserService
- Added email and password validation

### DIFF
--- a/src/main/java/mate/academy/service/AuthenticationService.java
+++ b/src/main/java/mate/academy/service/AuthenticationService.java
@@ -1,6 +1,9 @@
 package mate.academy.service;
 
+import mate.academy.model.User;
+
 public class AuthenticationService {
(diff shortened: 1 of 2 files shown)


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +16,14 @@ public class AuthenticationService {
...
-        return false;
+        User result = userService.findByEmail(email);
+        if (result == null) {
+            return false;
+        }
+        if (result.getPassword().equals(password)) {
+            return true;
+        } else {
+            return false;
+        }
```
```suggestion
        return user != null && user.getPassword().equals(password);
```

Elena-Bruyako (approved):
Good job
See comment how to simplify solution

PR #1311, opened 2026-06-17
added methods in AuthenticationService and UserService

Moskerito:
added methods in AuthenticationService and UserService


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
      * Return false in any other cases.
      */
     public boolean login(String email, String password) {
+        UserService userService = new UserService();
```
Elena-Bruyako: make UserService as class-level variable

src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -11,6 +13,11 @@ public class AuthenticationService {
...
+        User user = userService.findByEmail(email);
+        if (user != null) {
```

src/main/java/mate/academy/service/UserService.java
```diff
@@ -15,6 +15,14 @@ public class UserService {
      * Return <code>null</code> if there is no suitable user
      */
     public User findByEmail(String email) {
+
+        String emailCheck;
```
Elena-Bruyako: you don't need local variable here

Elena-Bruyako (changes requested):
good job
see comments

Elena-Bruyako (approved):
good job

PR #1312, opened 2026-06-18
changed login and findByEmail methods

drpacci:


src/main/java/mate/academy/service/AuthenticationService.java
```diff
@@ -10,7 +10,12 @@ public class AuthenticationService {
      * @return true if user by email exists and passed password is equal to user's password.
      * Return false in any other cases.
      */
+    private final UserService userService = new UserService();
+
     public boolean login(String email, String password) {
+        if (userService.findByEmail(email) != null) {
```
Elena-Bruyako: create local variable to avoid calling `userService.findByEmail(email)` twice

Elena-Bruyako (changes requested):
Good job
See comment

PR #1315, opened 2026-07-12
login service

mcDirek:


Elena-Bruyako (changes requested):
fix build first
revert all docs

PR #1318, opened 2026-07-27
add empty file.txt

sntmxm:


Elena-Bruyako (changes requested):
there are nothing to review, please provide HW implementation

